PR #63, opened 2026-05-23
build: add uv.lock for reproducible Docker builds

GeiserX:
## Summary
- Migrate Docker builds from `pip install -r requirements.txt` to `uv sync --frozen` with a committed lockfile
- Add all dependencies (merged from both requirements files) to `pyproject.toml` under `[project] dependencies`
- Both Dockerfiles now use `ghcr.io/astral-sh/uv:0.11` multi-stage pattern with deterministic installs
- Keep `requirements.txt` / `requirements-worker.txt` for backward compatibility with local dev

## Why
Every `docker build` previously resolved dependency versions at build time, making builds non-reproducible. The lockfile pins exact versions so CI and production always get identical environments.

## Dependabot compatibility
Dependabot supports `uv.lock` natively — it will open PRs updating both `pyproject.toml` bounds and lockfile pins.

## Test plan
- [ ] CI passes (lint, tests)
- [ ] Docker build succeeds for both `Dockerfile` and `Dockerfile.worker`
- [ ] `uv lock --check` confirms lockfile is in sync with pyproject.toml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,16 +5,15 @@ WORKDIR /build
 
 RUN apk add --no-cache gcc musl-dev libffi-dev
 
-COPY requirements.txt .
-RUN python -m venv /opt/venv
-ENV PATH="/opt/venv/bin:$PATH"
-RUN --mount=type=cache,target=/root/.cache/pip \
-    pip install --no-compile -r requirements.txt \
-    && find /opt/venv -type d -name "__pycache__" -exec rm -rf {} + 2>/dev/null; \
-       find /opt/venv -type f -name "*.pyc" -delete 2>/dev/null; \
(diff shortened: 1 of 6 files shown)


GeiserX:
Closing as superseded. This PR's goal — adding `uv.lock` for reproducible Docker builds — was already accomplished by #64 (merged the same afternoon, commit bc9ad3c). Main has shipped with `uv.lock` ever since, and the release workflow now runs `uv sync --frozen` against it.

Since then main has drifted, leaving this branch with conflicts in `pyproject.toml`, `requirements*.txt`, the Dockerfiles, and `uv.lock` itself. Rebasing would just re-resolve dependency pins that main already settled a different way, for no new value.

Closing rather than merging. Thanks!

PR #100, opened 2026-07-10
perf+security+test: deferred audit follow-ups (health growth, first-run token, orchestrator coverage)

GeiserX:
Implements three items deferred in #98, each independently verified (`ruff check` + `ruff format --check` + full `pytest` **1075 passed**). (Re-opened as a fresh PR — the original #99 auto-closed when its stacked base branch was deleted on merge; identical content, now rebased directly on `main`.)

## 1. Bound `health_events` growth ([PERF])
`check_ok`/`check_down` uptime samples (one per service every 5 min) dominate the table, but `get_health_scores` only reads a bounded window (`/api/health/scores` caps at 90 days) while retention kept everything 400 days.
- Trim samples to `HEALTH_CHECK_RETENTION_DAYS=95` (just past the 90-day max query, so **no allowed query can out-range its own samples** — uptime% unchanged) while keeping rare lifecycle events (start/stop/restart/crash) the full 400 days.
- Add a **weekly `VACUUM`** job (commit-first, `wal_checkpoint(TRUNCATE)`) — SQLite never shrinks the file on `DELETE` alone.

## 2. First-run setup token + trusted-proxy client IP ([SEC])
The "private network only" guard on first-run `/register` trusts `request.client.host`, which behind a reverse proxy is the *proxy* (loopback/private) — so it always passes and the first public visitor could seize the owner account.
- While no users exist, startup generates a one-time **setup token**, persists it (encrypted config), and logs it; `/register` now requires it (query param / `X-Setup-Token` header / form field) and it is retired the instant the owner is created.
- Resolve the real client IP behind a trusted proxy (opt-in `CASHPILOT_TRUSTED_PROXY`, right-most XFF) — used by the first-run network check **and** the login rate limiter, which previously collapsed every proxied client into one bucket keyed by the proxy IP.

## 3. orchestrator/worker_api coverage ([TEST])
Two lowest-covered, most Docker-privileged modules: **orchestrator.py 38%→73%**, **worker_api.py 44%→74%** (61 tests, Docker SDK fully mocked). Plus a `_login_attempts`/setup-token autouse reset fixture removing latent order-dependent flake.

Verified: `ruff` (check + format) clean; full `pytest` **1075 passed**.

## Summary by CodeRabbit

* **Security**
  * Added first-run setup-token protection for creating the initial account.
  * Restricted first-run access to private or loopback networks.
  * Improved client IP detection when operating behind a trusted proxy.
  * Setup tokens are cleared after successful initial registration.

* **Maintenance**
  * Added a weekly database maintenance task.

* **Reliability**
  * Expanded automated coverage for authentication, deployment, container management, and setup-token behavior.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -1073,13 +1073,27 @@ async def get_health_scores(days: int = 7) -> list[dict[str, Any]]:
 # --- Data Retention ---
 
 RETENTION_DAYS = 400
+# High-frequency uptime samples (check_ok / check_down, one per service every 5
+# minutes) are the dominant source of health_events growth, yet get_health_scores
+# only ever reads a bounded window. /api/health/scores caps that window at 90 days,
+# so we keep samples just past that (95d) — enough that no allowed query can
+# out-range its own samples, while still cutting the bulk sample rows ~76% versus
+# the 400-day lifecycle-event history (start/stop/restart/crash), which we keep in
+# full because those rows are rare and worth the long tail.
+HEALTH_CHECK_RETENTION_DAYS = 95
+_HEALTH_CHECK_EVENTS = ("check_ok", "check_down")
 
 
 async def purge_old_data() -> int:
-    """Delete earnings and health_events older than RETENTION_DAYS. Returns rows deleted."""
+    """Delete data past retention. Returns rows deleted.
+
+    Earnings and lifecycle health events are kept RETENTION_DAYS; the far more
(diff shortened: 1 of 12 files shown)


GeiserX:
@coderabbitai review

The query-param setup-token finding is addressed in f78c323: `_require_first_run_access` no longer reads `request.query_params`, the token is a form field / `X-Setup-Token` header only, the GET `/register` page renders behind just the network check, and the startup log no longer embeds the token in a URL.

PR #106, opened 2026-07-17
fix(auth): durably revoke sessions on user delete/demote (survives restart)

GeiserX:
Closes the highest-severity finding from the security audit (`CashPilot-rqw`).

## The bug
Deleting or demoting a user only bumped the **in-memory** `auth._USER_PWD_EPOCH` cache, and startup warm-up restored epochs solely from `users.password_changed_at`. So after **any UI restart** (webhook GitOps deploy, crash, reboot):
- a **deleted** user's row is gone → warm-up can't restore the epoch → it resets to 0 → their still-valid 30-day signed cookie is accepted again;
- a **demoted** user's `password_changed_at` was never touched → epoch not restored → their old cookie is accepted;

and because `require_role` reads the role from the **token claim** with no DB re-fetch, a removed/demoted owner regains **owner** (create users, deploy socket-privileged containers = host RCE, reveal the fleet key) for up to 30 days.

## The fix (surgical; the fast no-DB-in-request-path guard is unchanged)
- New `session_revocations(user_id, revoked_before)` table — **deliberately no FK to `users`**, so a deleted user's revocation outlives the row.
- `database.revoke_user_sessions()` persists it (monotonic — an older timestamp can't lower an existing revocation; prunes rows whose 30-day window has elapsed, since those tokens are already expired) + `list_session_revocations()` reads them back.
- The delete/demote routes now persist the revocation durably **in addition to** the in-memory epoch bump (immediate effect + restart-proof).
- Startup warm-up (factored into `_warm_session_epochs()`) restores the epoch cache from **both** `password_changed_at` and `session_revocations`, taking the later timestamp per user. Freshly-minted tokens (issued at login with the current DB role) are unaffected — a demoted user simply re-logs in and gets their new role.

Why revocation-epoch rather than DB role re-fetch: it keeps the deliberate "no DB call in the request path" design (an explicit comment in `decode_session_token`); the demoted/deleted user's old token is rejected outright, so the stale role in the token is never honored.

## Verification
`ruff check .` + `ruff format --check .` clean; **1149 tests pass**. New `tests/test_session_revocation.py`: durable store (persist / monotonic / prune / **survives user deletion**) + the restart restore (**revoke → clear cache → warm-up → old cookie rejected, fresh cookie accepted with the new role**) + password/revocation merge.

Forward-only schema (`CREATE TABLE IF NOT EXISTS` in `_SCHEMA`) — existing DBs get the table on next start. Not auto-merging: this touches the session-auth path and adds a table, so flagging for your review.

## Summary by CodeRabbit

* **Security**
  * Session cookies are now persistently revoked when an account is deleted or demoted, not just in memory.
  * Revocations survive restarts by being restored into the session-epoch cache during startup warm-up.
  * Session invalidation now merges revocation timing with password-change updates for correct precedence.

* **Bug Fixes**
  * Fixed cases where prior session revocations could be re-accepted after restart.
  * Expired revocation records are automatically pruned over time.

* **Tests**
  * Added regression coverage for durable revocation storage and startup warm-up behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to CashPilot are documented here.
 - **ProxyBase migrated to the current client** (#103). ProxyBase retired its Docker Hub image and old GHCR org and moved to `proxybase.org`, so the catalog entry no longer worked. The image is now `ghcr.io/proxybaseorg/peer-cli` (digest-pinned, multi-arch amd64/arm64/armv7 — arm64/Raspberry Pi now supported), credentials are the client's current `ID` (relabelled **Access Token**, masked) and `NAME` env vars, every URL points at `proxybase.org`, and datacenter IPs are now marked as accepted (residential still earns most). Existing ProxyBase deployments must be re-deployed with a fresh Access Token — see the [updated guide](docs/guides/proxybase.md)
 
 ### Security
+- Deleting or demoting a user now durably revokes their outstanding session cookies. Previously the revocation lived only in memory, so after a UI restart (deploy, crash, reboot) a deleted or demoted account's still-valid 30-day cookie was honored again with its old role. Revocations are persisted in a `session_revocations` table (which outlives the deleted user row) and restored into the session-epoch cache at startup
 - Write-only secrets: `GET /api/config` and `/api/env-info` no longer return stored credential values — only a set/not-set indicator. `CASHPILOT_SECRET_KEY` is never sent to the browser
 - Fleet key no longer sent on page load — revealed only via explicit owner-only action (`POST /api/fleet/api-key/reveal`)
 - Changing a password invalidates that user's existing sessions via a per-user epoch; the changer stays logged in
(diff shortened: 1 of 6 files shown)


GeiserX:
@coderabbitai review

GeiserX:
Thanks @coderabbitai — both addressed in the latest commit:

1. **Mutation/revocation ordering (Major)** — reordered both routes to persist the session revocation **before** the user mutation (`revoke_user_sessions` → then `update_user_role`/`delete_user`). Since the revocation is keyed by uid and has no FK to `users`, persisting it first is valid, and it eliminates the window where the mutation has committed but the old privileged cookie is still honored — a failure now leaves the account revoked-but-not-mutated (safe: logged out, re-login re-mints with the current role), never mutated-but-not-revoked. Chose reorder over a shared transaction because the two helpers each borrow the pooled connection and commit independently; revoke-first gives the same safety guarantee more simply.
2. **Deletion-survival test (Minor)** — the test now `create_user`s a real row before deleting it, so it genuinely verifies `session_revocations` is not dropped with the user (and would catch a future cascading FK).

PR #117, opened 2026-07-18
feat: detect deployed-image vs catalog-image drift and warn (CashPilot-5wi)

GeiserX:
## What

Detects when a running container's image no longer matches its catalog entry and surfaces an **update available** badge in the dashboard.

When a provider changes or re-pins its Docker image (the ProxyBase `proxybase/proxybase` → `ghcr.io/proxybaseorg/peer-cli` migration is the motivating case), a deployed container keeps reporting healthy while silently running a retired image. The operator has no signal to re-deploy.

## How

- `_split_image(ref)` — parses `repo`/`tag`/`digest` from a Docker ref (a `:` after the last `/` is a tag; before it is a registry port).
- `_image_outdated(deployed, catalog_image)` — True when the repo path changed, or both sides are digest-pinned to *different* digests. **Conservative:** empty/unknown images and a pure tag-vs-digest difference of the same repo are not flagged.
- `/api/services/deployed` now returns `image_outdated` per container-backed service.
- `app.js` renders a small `update available` badge (with an explanatory tooltip) in the status cell.

## Tests

`TestImageDrift` covers `_split_image` (digest, tag, bare, registry-port) and `_image_outdated` (repo migration, digest re-pin, identical, empty, tag-vs-digest). Full suite: 1164 passing, ruff clean.

Closes bead CashPilot-5wi.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -565,6 +565,37 @@ def _collector_needs_setup(slug: str, config: dict[str, str]) -> bool:
     return False
 
 
+def _split_image(ref: str) -> tuple[str, str, str]:
+    """Split a Docker image reference into (repository, tag, digest)."""
+    digest = ""
+    if "@" in ref:
+        ref, digest = ref.split("@", 1)
+    repo, tag = ref, ""
+    # A ':' is a tag only if it comes after the last '/', else it's a registry port.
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #118, opened 2026-07-18
fix: stable worker client_id instead of mutable hostname (CashPilot-ng1)

GeiserX:
## The bug (audit HIGH — worker_api.py:130, main.py:1751)

The bundled worker never sent a `client_id`, so the UI derived identity from the worker's **name**, which defaults to `socket.gethostname()`. Consequences:

- **Hostname collision:** two hosts sharing a common default hostname (`ubuntu`, `raspberrypi`, `docker-desktop`) collapse onto one identity. Once host A enrolls and gets its per-worker key, host B's shared-key heartbeats are `401`'d forever.
- **Rename lockout:** changing `CASHPILOT_WORKER_NAME` mints a brand-new identity, but the worker still presents its *old* persisted per-worker key → the UI rejects it → lockout.

## The fix

- Generate and persist a stable `client_id` in `/data/.worker_id` (random `uuid4().hex`), send it in the heartbeat, and key identity on it. **Name is now display-only.**
- Also surfaced on `/api/status` and the worker's config page for operator visibility.

## Migration (why it's safe to ship worker + UI together)

The auth path means a naive "always mint a UUID" would **lock out the existing fleet**: an already-enrolled worker holds its own key `K`, but a fresh UUID has no key on the UI and the worker never falls back to the shared key → permanent `401`. So:

- **Already-enrolled worker** (has `/data/.worker_key` but no `.worker_id`) → seeds `.worker_id` = its `WORKER_NAME`, i.e. the identity the UI already knows it by. No orphan row, no re-enrollment.
- **Brand-new worker** (no key) → random `uuid4().hex`.
- The UI already keys on `client_id` with a `name` fallback (`cid = body.client_id or body.name`), so old-worker/new-UI and new-worker/old-UI both interoperate.

## Tests

New `TestClientId` (6 cases): existing-id-file read, random mint for a brand-new worker (≠ display name), **migration preserves WORKER_NAME identity for an enrolled worker**, stability across calls, persist-failure fallback to an in-memory id, and the heartbeat payload carrying `client_id` while `name` stays present. Full suite: 1168 passing, ruff clean.

Closes bead CashPilot-ng1.

### DIFF
--- a/app/worker_api.py
+++ b/app/worker_api.py
@@ -20,6 +20,7 @@ import os
 import platform
 import re
 import socket
+import uuid
 from contextlib import asynccontextmanager
 from datetime import UTC, datetime
 from html import escape as _esc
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #119, opened 2026-07-18
fix+refactor: api_worker_command earnings bug + main.py de-duplication (CashPilot-1k5)

GeiserX:
Behavior-preserving refactor of `app/main.py`, plus a real correctness fix. Each step is its own commit; the full suite (1172) is green after every step.

## The bug (headline) — `api_worker_command` skipped all bookkeeping

`POST /api/workers/{id}/command` proxied the Docker action but ran **none** of the side-effects the primary lifecycle routes do. A **deploy** through it never created a deployments row — so the service **earned $0 forever** — and a **remove** leaked the row (health/metrics also never recorded). Fixed by mirroring the canonical routes' bookkeeping (`save_deployment`/`remove_deployment`/`record_health_event`/`metrics`/`_run_collection`), keyed on the command, **on a successful proxy only**. Owner-for-deploy / writer-else gating is unchanged.

## De-duplication (all behavior-preserving)

1. **One worker proxy.** `_proxy_worker_command/_deploy/_logs` and `api_worker_command`'s inline dispatch all repeated the fetch-worker → verified-URL+auth → httpx → error-map sequence. Consolidated into `_proxy_to_worker(worker_id, method, path, *, json, params, timeout)`. Keeps the **concrete httpx verbs** (never `client.request`) so per-verb test mocks still land; callers keep their timeouts (deploy 60s, else 30s) and params (logs clamp, remove's `delete_volumes`).
2. **One lifecycle trio.** The flat `/api/{stop,restart,remove}/{slug}` and service-scoped `/api/services/{slug}/...` routes had byte-identical bodies → both delegate to `_svc_stop/_svc_restart/_svc_remove`. All 9 routes, paths, and `_require_writer`/`_require_owner` guards unchanged.
3. **One entry-meta tail.** `api_services_deployed`'s container and external loops repeated the identical cashout/referral/website tail → `_apply_service_meta(entry, svc)`. Key insertion order (and the JSON) unchanged.

## Tests

`api_worker_command` now has command coverage for deploy/stop/restart/start/remove asserting each records its bookkeeping (main.py 98%, total 94%, floor 90%). Full suite **1172 passing**; `ruff check` + `ruff format --check` clean.

## Deliberately deferred (tracked as a follow-up bead)

Per architect review, this PR does **not** split `main.py` into new route modules or resolve the benign `main→routers→main` import cycle. Splitting now would break ~40 `app.main.*` patch seams (`httpx.AsyncClient`×25, `socket.getaddrinfo`×7, `FLEET_API_KEY`×34, `_resolve_worker_id`, `database.*`/`catalog.*`) and widen the cycle — high interface risk for low correctness value. The god-module is already materially reduced by the three dedups above; the pure structural split is a focused follow-up.

Closes bead CashPilot-1k5.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1153,64 +1153,63 @@ async def _get_verified_worker_url(worker: dict[str, Any]) -> tuple[str, dict[st
     return url, headers
 
 
+async def _proxy_to_worker(
+    worker_id: int,
+    method: str,
+    path: str,
+    *,
+    json: dict[str, Any] | None = None,
+    params: dict[str, str] | None = None,
+    timeout: float = 30,
+) -> dict[str, Any]:
+    """Proxy one request to a worker's REST API and return its parsed JSON.
+
+    Single home for the fetch-worker -> verified-URL+auth-header -> httpx call ->
+    error-mapping sequence the deploy/command/logs paths all repeated. Dispatch uses
+    the concrete httpx verbs (never client.request) so per-verb test mocks keep
+    landing; the caller owns the timeout (deploy needs 60s, the rest 30s) and any
+    query params (the logs line clamp, remove's delete_volumes).
+    """
+    worker = await database.get_worker(worker_id)
+    url, headers = await _get_verified_worker_url(worker)
+
+    try:
+        async with httpx.AsyncClient(timeout=timeout) as client:
+            verb = method.upper()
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #120, opened 2026-07-18
fix(ui): unify deploy flows so the detail view validates + surfaces errors (CashPilot-cyc, part 1)

GeiserX:
## The bug

The catalog/detail view's `deployServiceToWorkers` was a **weaker duplicate** of the setup wizard's `deployService` (audit: app.js:1837). It:
- skipped **required-field validation**,
- **swallowed per-worker server errors silently** — a failed deploy looked successful to the user,
- used a non-slug-scoped checkbox class (`.deploy-worker-cb` vs `.setup-deploy-worker-cb[data-slug]`),
- never updated wizard state.

## The fix

Extract the proven `deployService` body into a shared **`_deployToWorkers(slug, workerIds)`** (validation → env collection → per-worker POST → per-worker error toasts → status + wizard state), and make **both** entry points thin wrappers over it with consistent slug-scoped selectors. Added `data-slug` to the detail-view checkbox so its selector matches; the env-collection `[data-key]` filter keeps the worker checkboxes out of the env payload.

## Scope

This lands the **duplicate-deploy-flow** part of `CashPilot-cyc`. The larger HTML-builder componentization (`renderServiceRow`, the setup/detail form builders) and inline-`onclick` removal remain open — they change rendered markup and every handler's wiring, so they need **browser verification against a running UI** (and feed directly into the CSP-tightening bead `guw`). The bead stays open for that.

## Verification

- `node --check` clean.
- `deployService` is behavior-identical (verbatim body extraction, unchanged selector).
- `deployServiceToWorkers`' new slug-scoped selectors are backed by the rendered `data-slug`/`data-key` attributes (static analysis).
- Full Python suite still 1172 green.
- *Browser smoke test was not possible in this environment (Chrome extension unavailable); verification is static + syntax.*

Partial progress on bead CashPilot-cyc.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service deployment validation across the Setup Wizard and Service Detail views.
  * Added clearer progress indicators and per-worker failure notifications during deployment.
  * Fixed worker selection to ensure deployments target the correct workers.
  * Deployment status now updates consistently after successful operations.

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -1491,18 +1491,20 @@ const CP = (() => {
     </div>`;
   }
 
-  async function deployService(slug) {
+  // Deploy `slug` to the given worker node ids: collect the env inputs, validate the
+  // required fields, POST per worker, and surface each per-worker failure. Shared by both
+  // deploy entry points (the setup wizard and the catalog/detail view) so validation and
+  // error reporting are identical — the detail view previously skipped validation and
+  // swallowed server errors silently.
+  async function _deployToWorkers(slug, workerIds) {
     const statusEl = document.getElementById(`deploy-status-${slug}`);
-    const checkboxes = document.querySelectorAll(`.setup-deploy-worker-cb[data-slug="${slug}"]:checked:not(:disabled)`);
-    const workerIds = Array.from(checkboxes).map(cb => parseInt(cb.dataset.wid));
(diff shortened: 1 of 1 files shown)


GeiserX:
@coderabbitai review

PR #122, opened 2026-07-18
fix(security): HSTS + CSP hardening + cookie Secure behind a trusted proxy

GeiserX:
Backend-only security hardening surfaced by the research security panel (items #2 session-cookie, #3 headers). **Keeps the current `'unsafe-inline'`** — dropping it is the separate `guw` refactor (which needs the inline-handler removal first).

## Changes

**CSP — additive hardening (zero UI impact today):**
- `base-uri 'none'` — blocks a `<base>` tag hijacking relative `/static/js/app.js`.
- `object-src 'none'` — no plugins.
- `form-action 'self'` — a form can't POST credentials off-origin if an XSS is ever found.

**HSTS — emitted only on real HTTPS:**
- `Strict-Transport-Security: max-age=63072000; includeSubDomains` when the request scheme is https, or `X-Forwarded-Proto: https` **when `CASHPILOT_TRUSTED_PROXY` is set**. Plain-HTTP local dev is never pinned (which would hard-break it), and an untrusted/spoofable header can't trigger it.

**Session cookie `Secure` in `auto` mode:**
- Previously `Secure` only turned on when `CASHPILOT_BASE_URL` started with `https`. On the documented Caddy topology, TLS terminates at the proxy and operators configure the domain *in Caddy*, so `CASHPILOT_BASE_URL` is unset → the 30-day session cookie shipped **non-Secure** over the proxy hop (LAN/MITM interceptable → owner takeover). Now `auto` also turns `Secure` on when `X-Forwarded-Proto: https` behind a trusted proxy. `set_session_cookie` gains an **optional** `request` arg (backward-compatible — callers without it keep the old behavior).

## Tests
New coverage: the three CSP directives; HSTS present/absent by scheme + proxy-trust; cookie `Secure` on/off by `X-Forwarded-Proto` + trust; explicit-override + no-request backward-compat. Full suite **1179 green**, ruff clean.

## Not in scope (tracked)
- Dropping `script-src 'unsafe-inline'` (nonce/delegation) → bead `guw`.
- CSRF Origin-check + POST `/logout` → follow-up (SameSite=Lax + no-CORS already mitigate today).

## Summary by CodeRabbit

* **Security Enhancements**
  * Improved session-cookie security when the application runs behind a trusted HTTPS reverse proxy.
  * Added stricter browser security policies, including stronger protection against framing, unsafe form submissions, and unwanted content sources.
  * Improved HSTS behavior so HTTPS enforcement correctly reflects trusted proxy configuration.

* **Bug Fixes**
  * Fixed authentication and registration flows to consistently apply secure session-cookie settings.
  * Preserved existing behavior for deployments that do not use proxy forwarding.

### DIFF
--- a/app/auth.py
+++ b/app/auth.py
@@ -181,22 +181,32 @@ _SECURE_COOKIE = os.getenv("CASHPILOT_SECURE_COOKIE", "auto").lower()
 _SECURE_COOKIE_TRUE = {"true", "1", "yes", "on"}
 _SECURE_COOKIE_FALSE = {"false", "0", "no", "off"}
 
+# Only trust an X-Forwarded-Proto header (to detect TLS terminated at a reverse proxy)
+# when the operator opts in — matches app.deps._TRUST_PROXY; read here to avoid importing
+# deps into auth (deps imports auth, so the reverse would be a cycle).
+_TRUST_PROXY = os.getenv("CASHPILOT_TRUSTED_PROXY", "").strip().lower() in ("1", "true", "yes", "on")
 
-def set_session_cookie(response: RedirectResponse, token: str) -> RedirectResponse:
+
+def set_session_cookie(response: RedirectResponse, token: str, request=None) -> RedirectResponse:
     # Secure-flag precedence (highest wins):
     #   1. CASHPILOT_SECURE_COOKIE explicitly truthy -> Secure on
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #123, opened 2026-07-18
perf: batch health-check writes + PRAGMA synchronous=NORMAL + concurrent exchange fetch

GeiserX:
Three low-risk hot-path fixes from the research performance panel.

## 1. Batch the health-check writes (top finding)
`_run_health_check` committed **one health event per deployed service** — up to ~49 fsync'd commits per cycle × 288 cycles/day. Now it collects the cycle's events and writes them in **one transaction** via a new `database.record_health_events` (one `executemany` + one `commit`) → **27–98× fewer commits** on the write-heaviest path. Worsens with both fleet size and concurrent viewers on the single shared aiosqlite connection, so this is the highest-impact one.

## 2. `PRAGMA synchronous=NORMAL`
In WAL mode NORMAL is durable across app crashes (only power loss can drop the last txns) and **skips an fsync on every commit** — a fleet-wide win on top of #1.

## 3. Concurrent exchange-rate fetch
`refresh()` fetched CoinGecko then Frankfurter **sequentially**, and it's `await`ed on the **startup critical path** — up to 30s worst case (two 15s timeouts). Now both run concurrently via `asyncio.gather(..., return_exceptions=True)`: halves the worst case, and one source's network failure no longer discards the other's success (each still tracks its own staleness / non-200 internally).

## Tests
- New `record_health_events` batched-write + empty-no-op test.
- Health-check tests updated to the batched contract (assert on the single event list).
- Full suite **1173 green**, ruff + format clean.

## Deferred (also from the panel, lower priority)
- Decouple container-stats refresh from the 60s heartbeat / parallelize `_collect_stats`.
- Short-TTL cache for `_get_all_worker_containers`/`get_config` (dashboard fires 15 round-trips, 3 datasets refetched 2–3×).
- Per-collector `asyncio.wait_for` ceiling; covering index on `health_events`.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -280,6 +280,10 @@ async def _get_db() -> _BorrowedConnection:
         await conn.execute("PRAGMA journal_mode=WAL")
         await conn.execute("PRAGMA foreign_keys=ON")
         await conn.execute("PRAGMA busy_timeout=5000")
+        # In WAL mode NORMAL is durable across app crashes (only a power loss can lose
+        # the last transactions) and skips an fsync on every commit — a large win on the
+        # write-heavy health-check path that commits per service each cycle.
+        await conn.execute("PRAGMA synchronous=NORMAL")
         _shared_conns[key] = conn
 
     return _BorrowedConnection(conn)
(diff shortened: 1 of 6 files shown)


GeiserX:
@coderabbitai review

PR #124, opened 2026-07-18
refactor(ui): extract shared action-button icon constants (CashPilot-cyc)

GeiserX:
First safe step of the `cyc` builder componentization, now that browser verification is unblocked.

## What
`renderServiceRow` duplicated the restart/stop/logs SVG icons verbatim — 6 copies (3 icons × single-instance row + per-node sub-row). Extracted to `ICON_RESTART`/`ICON_STOP`/`ICON_LOGS` module consts.

## Why it's provably safe
Per the migration review, "byte-identical" is only guaranteed at the **DOM** level. Verified three ways:
1. **DOM golden-master** — rendered the multi-instance row + expanded per-node sub-rows (via a mocked running service in a headless Chromium harness), dumped the table `innerHTML` before and after: **byte-identical (10346 bytes, zero diff)**.
2. `node --check` clean; the `CP` export surface is unchanged.
3. Playwright harness: the row + all action buttons (stop/restart/logs ×2, credential, expand) render and the expand toggle works, with **0 console/page errors**.

## Scope
Just the icon consts (the truly-identical, single-line strings — zero indentation risk). The env-field / worker-checkbox helpers (which live in the wizard + detail views and differ by indentation/class) follow as separate golden-master-verified PRs, ahead of the `guw` CSP work.

Part of bead CashPilot-cyc.

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -425,6 +425,12 @@ const CP = (() => {
     }
   }
 
+  // Action-button icons, extracted so the single-instance row and the per-node sub-row
+  // share one copy each (cyc componentization). Byte-identical to the inline SVGs replaced.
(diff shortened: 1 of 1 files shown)


GeiserX:
@coderabbitai review

PR #127, opened 2026-07-31
fix: refuse deploying broken/dropped services, not just dead (CashPilot-rp3)

GeiserX:
The catalog listing (`/api/services/available`) hides services with status `broken`, `dead` or `dropped`, but the deploy gate only refused `dead`. A direct link, a stale page or an API client could therefore still deploy a service the catalog already knows is non-functional — which then silently earns nothing and becomes a support issue.

**Fix:** both paths now share one `_UNDEPLOYABLE_STATUSES` constant, so they cannot drift apart again. `broken` returns **409** (may come back), `dead`/`dropped` return **410** (gone). The existing "no longer available" wording is preserved.

**Tests:** new cases for broken (409) and dropped (410); the existing dead-service test is unchanged. Full suite **1182 passing**, ruff + format clean.

Closes bead CashPilot-rp3.

## Summary by CodeRabbit

- **New Features**
  - Improved service availability filtering to exclude services that cannot be deployed.
  - Deployment requests now provide status-specific responses for broken, dead, and dropped services.

- **Bug Fixes**
  - Prevented deployment attempts for broken and dropped services, ensuring unavailable services are handled consistently.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -587,6 +587,14 @@ def _collector_needs_setup(slug: str, config: dict[str, str]) -> bool:
     return False
 
 
+# Catalog statuses a service must never be deployed with. Shared by the catalog
+# listing (which hides them) and the deploy gate (which refuses them) so the two
+# can't drift apart again — previously the listing hid all three but deploy only
+# refused "dead", so a direct link or a stale page could still deploy a broken or
+# dropped service and then silently earn nothing.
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

GeiserX:
Fixed in the latest commit — the stop test now asserts the proxied call reaches `/api/containers/speedshare/stop`, since a bare 200 would also pass if the route did nothing. Thanks!

PR #128, opened 2026-07-31
fix: let operators opt specific volume roots past the block (CashPilot-52w)

GeiserX:
## The bug

The worker refuses bind mounts under system roots — including `/mnt`, to stop a fleet-key holder mounting a co-located app's secrets off the Unraid array. But **Storj** needs a large data directory, and on Unraid the only such path is `/mnt/user/...`, so its deploy 403s: **Storj is undeployable through the worker on the platform most CashPilot users run**. The practical effect is users deploying it by hand outside CashPilot, which is strictly worse for security than a scoped exception.

## The fix

Deny-by-default is unchanged. `CASHPILOT_ALLOWED_VOLUME_ROOTS` lets the operator name the exact directories they accept, colon-separated:

```
CASHPILOT_ALLOWED_VOLUME_ROOTS=/mnt/user/storj
```

**Only a specific subdirectory can be opted in.** An entry that is itself a whole system root (`/`, `/mnt`, `/var/run`, `/etc`, ...) is ignored with a warning, so a careless value can never re-expose `/var/run/docker.sock`, `/etc` or an entire array. Prefix matching is path-aware, so `/mnt/user/storj` does not allow `/mnt/user/storjevil`.

## Tests

9 new cases: the allowlisted subdir passes; siblings on the array stay blocked; the `storjevil` prefix trap; each whole-system-root entry is refused (parametrised); multi-entry parsing. Full suite **1189 passing**, ruff clean. Documented in `docs/fleet.md`.

Closes bead CashPilot-52w.

### DIFF
--- a/app/worker_api.py
+++ b/app/worker_api.py
@@ -420,6 +420,31 @@ _BLOCKED_VOLUME_ROOTS = {
 }
 
 
+# System roots that may never be opened up, at ANY depth. An allowlist entry under
+# one of these would hand a third-party container the Docker socket (/run,
+# /var/run), the host's secrets (/etc, /root) or its devices (/dev) — so unlike the
+# data roots below, no subdirectory of these is ever acceptable.
+_NEVER_ALLOWLISTABLE = (
+    "/etc",
+    "/root",
+    "/proc",
+    "/sys",
+    "/dev",
+    "/boot",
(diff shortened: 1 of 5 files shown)


GeiserX:
@coderabbitai review

PR #129, opened 2026-07-31
fix(security): drop all capabilities + no-new-privileges on deployed containers (CashPilot-a5p)

GeiserX:
## The gap

Service containers are **third-party, closed-source proxyware images**, but they were deployed with in-container **root plus Docker's full default capability set** — `NET_RAW` (ARP/DNS spoofing on the container bridge), `MKNOD`, `SETUID`, `SETGID`, `DAC_OVERRIDE`, `SYS_CHROOT`. `deploy_raw` never set `cap_drop`, `security_opt` or `pids_limit`.

The hardening was **inverted**: `docker-compose.yml` already sets `no-new-privileges` on the *trusted* UI and worker, while the *untrusted* containers got nothing.

## The change

- `cap_drop: ["ALL"]`, then re-add only the capabilities that service's own catalog entry declares. Today that union is exactly **`NET_ADMIN` for mysterium**; the other 48 declare none and are plain outbound TCP clients.
- `security_opt: ["no-new-privileges:true"]`
- `pids_limit` (`CASHPILOT_PIDS_LIMIT`, default 512) so a runaway or hostile image can't exhaust the host PID namespace.
- **`privileged` removed from `deploy_raw` entirely.** The worker already refused it at spec validation; now the dangerous state is unrepresentable rather than merely unreached.

This is the practical version of "sandbox the containers" — it removes real attack surface with near-zero breakage, unlike a custom OCI runtime (which doesn't install cleanly on Unraid, needs a custom kernel on Pi, is redundant under Docker Desktop's VM on macOS, costs ~1.7× network throughput on a network-bound workload, and would break mysterium's host networking anyway).

## Verified against real images, not assumed

The one risky assumption was that dropping `CHOWN`/`SETUID` would break entrypoints that chown a data dir or drop privileges. Tested locally with Docker:

| image | baseline | with `--cap-drop ALL --security-opt no-new-privileges --pids-limit 512` |
|---|---|---|
| `honeygain/honeygain` | exits 1, usage banner (no credentials) | **identical** |
| `storjlabs/storagenode` | supervisord up, storagenode exits 1 (no identity) | **identical** |

Zero permission-related log lines in either hardened run. *Caveat:* neither got far enough to write into its data dir (no credentials/identity), so the deep chown path is covered by reasoning — only mysterium declares a capability — rather than by execution.

## Tests

3 new: hardening flags always applied; a catalog-declared capability is still added back on top of the drop; `privileged` is no longer an accepted argument (`TypeError`). Full suite **1183 passing**, ruff + format clean. Documented in `docs/fleet.md`.

Closes bead CashPilot-a5p.


GeiserX:
@coderabbitai review

PR #130, opened 2026-07-31
fix: record the FX rate with each earnings reading (CashPilot-rrr)

GeiserX:
## Why this is urgent rather than merely missing

The `earnings` table stores `(platform, balance, currency, date)` and **no exchange rate**, while `exchange_rates.py` caches only *current* rates. So the USD value of a historical non-USD reading (MYST, GRASS, ...) cannot be reconstructed afterwards at any accuracy.

Unlike most gaps, **this one cannot be fixed retroactively** — every collection cycle that runs without the column loses that data permanently. That's why it's worth landing before the features that consume it.

## Change

- Nullable `fx_rate_usd` column on `earnings`, plus an idempotent `ALTER TABLE` migration for existing databases (matching the existing `workers`/`users` migration pattern).
- Populated at collection time from `exchange_rates.to_usd(1.0, currency)` — the currency→USD rate, `1.0` for USD, and **`NULL` when the rate is unknown** so a missing rate is never silently stored as a wrong one.
- Existing rows keep `NULL` rather than being back-filled with today's (wrong) rate.

## Tests

5 new: the rate is stored with the reading; it updates on conflict; it defaults to NULL; the migration adds the column to a pre-existing DB **without losing its rows**; and the migration is idempotent. Full suite **1185 passing**, ruff + format clean.

Unblocks the payout ledger (`glc`), power/net-profit (`l01`) and any future tax/CSV export.

Closes bead CashPilot-rrr.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -106,11 +106,11 @@ CREATE TABLE IF NOT EXISTS earnings (
     balance    REAL    NOT NULL,
     currency   TEXT    NOT NULL DEFAULT 'USD',
     date       TEXT    NOT NULL,
-    -- Currency -> USD rate at the moment this reading was taken. Rates are only
-    -- cached live, so without storing it here the historical value of a non-USD
-    -- balance (MYST, GRASS, ...) cannot be reconstructed later at any accuracy —
-    -- which is what a net-profit or tax export needs. NULL for USD or when the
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

Issue #103, opened 2026-07-17
[Bug]: ProxyBase doesnt work, devs updated backend


kissofkill:
### Description

User ID is Peer ID now; also, they have access token for pairing devices

and new docker is
docker run -d ghcr.io/proxybaseorg/peer-cli:latest XXX XXX
instead of
docker run -d ghcr.io/proxybase-org-company/peer-cli:latest XXX XXX

### Steps to Reproduce

title

### Expected Behavior

-

### Deployment Method

Docker Compose

### CashPilot Version

1

### Architecture

arm64 (aarch64 / Raspberry Pi)

### Relevant Logs

```shell

```

### Checks

- [x] I have checked whether this is already fixed in the latest version
- [x] I searched existing issues and this is not a duplicate

wrenchyyy:
Hi! 👋

Representing ProxyBase here. First of all, thanks for including us in your list of bandwidth-sharing providers.

There are a few details that have changed or are currently inaccurate, and we'd appreciate it if they could be updated.

1. Docker image location

Our Docker image is now hosted on GitHub Container Registry (GHCR) instead of Docker Hub. Please update the documentation accordingly.

2. Incorrect information on guide

The guide at:

https://geiserx.github.io/CashPilot/guides/proxybase/

currently links to **proxybase.io**, which is not our service.

We are a registered business, and all of our official services are under the **proxybase.org** domain.

For users interested in bandwidth sharing, the appropriate URL is:

**https://peer.proxybase.org**

3. Requirements section

The guide currently states that we only accept residential IPs, which is not accurate.

We also accept datacenter IPs. While traffic for them is currently much lower than residential IPs, they are supported and are not banned from the platform.

4. Credential terminology

The documentation still refers to **"user-id"**. This has been replaced with **"Access Token"**, so updating that terminology would help avoid confusion for new users.


We also have an official Discord server where we post project updates and announcements. If you use Discord, we'd be happy to have you join us there as well. 🙂

Thanks
Wrenchy

GeiserX:
Thanks for the report, @kissofkill 🙏 — the new image path and the arm64/Raspberry Pi detail made this quick to track down (and you were right that the ID is issued as a pairing token now).

Root cause: the catalog still shipped ProxyBase's retired Docker Hub client from before they consolidated onto GHCR and proxybase.org — which is also why it never had a chance on your Pi, since that old image was amd64-only.

[...]

Issue #82, opened 2026-06-10
[Bug]: Repocket integration requests password instead of API key


jessy2027:
## Description

The CashPilot UI asks for a Repocket email and password.

However, the Repocket container logs indicate that the required credentials are:

- RP_EMAIL
- RP_API_KEY

Using the account password does not work, and using the API key in the password field also fails to authenticate.

## Expected Behavior

The UI should request:

- Email
- API Key

or clearly indicate that the password field expects a Repocket API key.

## Actual Behavior

The UI requests a password, which is misleading because the Repocket container expects an API key.

This makes the integration difficult to configure and may prevent successful authentication.

## Logs

Logs: repocket (worker #1)

2026-06-10T15:58:29.847570755Z peer-client-version: 1.1.45
2026-06-10T15:58:29.863119107Z Your current version is : 1.1.46
2026-06-10T15:58:29.863380284Z User credentials are missing!
2026-06-10T15:58:29.863390138Z 
2026-06-10T15:58:29.863394729Z Please declare environment variables RP_EMAIL and RP_API_KEY.
2026-06-10T15:58:29.863398513Z 
2026-06-10T15:58:29.863402037Z To get your Api Key please visit in https://app.repocket.co/
2026-06-10T15:58:29.863405591Z 
2026-06-10T15:58:29.863409038Z Alternatively, supply credentials as command line arguments:
2026-06-10T15:58:29.863412525Z -e 'Your Email' -p 'Your Api key'
2026-06-10T15:58:30.928952096Z peer-client-version: 1.1.45
2026-06-10T15:58:30.948254849Z Your current version is : 1.1.46
2026-06-10T15:58:30.948474519Z User credentials are missing!
2026-06-10T15:58:30.948479398Z 
2026-06-10T15:58:30.948481588Z Please declare environment variables RP_EMAIL and RP_API_KEY.
2026-06-10T15:58:30.948483489Z 
2026-06-10T15:58:30.948485189Z To get your Api Key please visit in https://app.repocket.co/
2026-06-10T15:58:30.948486985Z 
2026-06-10T15:58:30.948488637Z Alternatively, supply credentials as command line arguments:
2026-06-10T15:58:30.948498586Z -e 'Your Email' -p 'Your Api key'
2026-06-10T15:58:32.828163268Z peer-client-version: 1.1.45
2026-06-10T15:58:32.843908576Z Your current version is : 1.1.46
2026-06-10T15:58:32.844163394Z User credentials are missing!
[...]

Issue #59, opened 2026-05-22
BUG: use EARNFM_TOKEN instead of email/password for earnfm


jnofsinger:
### Description

services/bandwidth/earnfm.yml defines env keys EARNFM_EMAIL and
EARNFM_PASSWORD (Supabase-style auth flow), but the upstream
earnfm/earnfm-client image does not consume those variables. It
authenticates via a single EARNFM_TOKEN env var — a UUID API key
copied from the user's account settings at app.earn.fm.

With the YAML as-is, the container starts but immediately loops on:

    {"status":400,"message":"Missing Property","data":[{"field":"APIKey",
     "value":"","message":"required APIKey"}]}
    Connection is halted because: invalid API key

— because EARNFM_TOKEN is never set in the container environment, no
matter what the user enters in the setup wizard.

Replace the two env entries with a single required, secret EARNFM_TOKEN
entry, and update the description to direct users to copy their API
key from the EarnFM dashboard rather than entering account credentials.

Per upstream Docker Hub documentation, the canonical run command is:

    docker run -d --restart=always \
      -e EARNFM_TOKEN="<uuid-api-key>" \
      earnfm/earnfm-client:latest

Verified locally: container authenticates against EarnFM's API,
begins bandwidth sharing, and is tracked correctly by the CashPilot
earnings collector.

### Steps to Reproduce

main web url, click on deploy earnfm, then review docker logs for earnfm

### Expected Behavior

starts normally

### Deployment Method

Docker Compose

### CashPilot Version

v0.2.49

### Architecture

amd64 (x86_64)

### Relevant Logs

```shell

```

### Checks

- [x] I have checked whether this is already fixed in the latest version
- [x] I searched existing issues and this is not a duplicate

franko-c:
env block fix lines up with what upstream's docker hub page documents, but the wording in the same file still points users at the old flow — `description` (lines 6-9), `short_description` (line 10), and `collector.notes` near the bottom all reference Supabase email/password auth, so a complete swap touches four sections of `services/bandwidth/earnfm.yml`, not just `docker.env`:

```yaml
env:
  - key: EARNFM_TOKEN
    label: "API Token"
    required: true
    secret: true
    description: "UUID API key from app.earn.fm → Settings → API"
```

`secret: true` on the new entry is the right setting — a UUID API key is a credential the worker should redact in logs, not an identifier.

[...]

Issue #44, opened 2026-05-12
[Bug]: Presearch environment var is wrong.


codeshard:
### Description

When launching presearch thru the Service Catalog, the name of the environment variable `PRESEARCH_REGISTRATION_CODE` is wrong, must be `REGISTRATION_CODE`. Check https://docs.presearch.io/nodes/setup#generic-setup-steps

### Steps to Reproduce

1. Go to the Service Catalog of CashPilot.
2. Launch Presearch dePin App.
3. Check the logs of presearch in Deployed Services. 

### Expected Behavior

The env var of Presearch is setted up correctly, to REGISTRATION_CODE, so the bad registration code error is gone. 

### Deployment Method

Docker Compose

### CashPilot Version

v0.2.89

### Architecture

amd64 (x86_64)

### Relevant Logs

```shell
/app # env
HOSTNAME=cashpilot-presearch
SHLVL=1
HOME=/root
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PRESEARCH_REGISTRATION_CODE=11111111111111111111111111111111
PWD=/app



2026-05-12T09:30:00.551Z info: [gateway.presearch.com] Connecting...
2026-05-12T09:30:03.107Z error: Error response from gateway: Bad registration code: Please visit https://nodes.presearch.com to obtain a valid registration code
```

### Checks

- [x] I have checked whether this is already fixed in the latest version
- [x] I searched existing issues and this is not a duplicate

GeiserX:
Thanks for the clear bug report! Fixed in #45 — the env var is now correctly set to `REGISTRATION_CODE` per the Presearch docs. The fix will be available in the next patch release (auto-built shortly).

codeshard:
Thanks for the really fast fix! Amazing project btw! Kudos!

PR #131, opened 2026-07-31
feat: weekly catalog liveness check with referral-link guard (CashPilot-owv)

GeiserX:
## Why

Catalog rot is the most common user-facing failure in this project — a provider changes its backend, drops its image, or ends its referral programme, and **a user finds out before the maintainer does**. Roughly 42% of user-filed issues are stale catalog/collector entries (#103, #82, #59, #44, #24).

## What

`scripts/check_catalog_liveness.py` + a weekly workflow. Per service it checks the **website**, the **referral signup URL**, and the **docker image**, then keeps **one** rollup issue up to date (never a new issue per run).

Referral findings get their own section, called out first: a dead or code-stripped referral link is **direct lost revenue**, not just a broken link. This effectively turns the standing "verify my referral link is correctly set" rule into a scheduled check.

## Deliberately conservative, so the report stays worth reading

A job that is red every week is a job nobody reads. So:

- **5xx → `unreachable`, not `dead`** — a provider having a bad afternoon is not a retired service.
- **Registry rate-limit / auth-gated → `unreachable`** — reporting that as dead would send someone deleting a live service.
- **401/403/405/429 → alive** — bot guards prove the host answered.
- **HEAD redirect-loops retry with GET** — caught a live provider being misreported during testing.
- Services already marked `dead`/`dropped` are skipped (no noise for decisions already made).
- **A dead link never fails the run**; only the script failing to run does.

## Verified against the real catalog, not just mocked

Ran it over all 49 services locally — **11 findings**, including exactly the case it was built for:

> `proxylite` — **referral code lost**: `https://proxylite.ru/?r=KMUPRZIZ` now redirects to a bare `https://proxylite.ru/`.

plus unreachable hosts (`earncc`, `koii`, `teneo`, `wizardgain`, `spide`).

## Tests

32 offline tests covering the decision logic — status classification, the referral-collapse heuristic (including the cases that must **not** fire), rate-limit vs missing-manifest, YAML loading, dead-service skipping, and report shaping. No network or docker in the test suite. Full suite **1212 passing**; repo-wide `ruff check` + `ruff format --check` clean.

Public repo, so `ubuntu-latest` (free, ephemeral) per the CI runner policy.

Closes bead CashPilot-owv.

### DIFF
--- a/.github/workflows/catalog-liveness.yml
+++ b/.github/workflows/catalog-liveness.yml
@@ -1,0 +1,64 @@
+name: Catalog Liveness
+
+# Catalog rot is the most common user-facing failure in this project: a provider
+# changes its backend, drops its image or ends its referral programme, and a user
+# finds out before the maintainer does. This checks every service's external
+# references weekly and keeps ONE rollup issue up to date.
+
+on:
+  schedule:
+    - cron: "17 6 * * 1" # Mondays, 06:17 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: catalog-liveness
+  cancel-in-progress: false
+
+jobs:
+  check:
+    # Public repo -> GitHub-hosted runners are free and ephemeral.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #133, opened 2026-07-31
fix: stop reporting a session-capture referral link as dead

GeiserX:
The weekly catalog-liveness check reported ProxyLite's referral link as `referral code lost` and counted it as a problem. Verified against the live site — that is wrong:

| Request | Response |
|---|---|
| bare homepage | `200` |
| `?r=KMUPRZIZ` | `302` + `Set-Cookie: PHPSESSID` → `/` |
| `?r=ZZZZNOPE9` (bogus) | `302` → `/` |

The site special-cases `?r=`, captures the code into the PHP session and redirects to a clean URL. That is a **working** referral link, and the ProxyLite node is live (container relaying traffic continuously, `restarts=0`).

Distinguishing capture-and-redirect from a genuinely dropped code requires an account, so this is now reported as **inconclusive** rather than dead: still listed under 'Could not verify', but not counted in the problem tally that opens the weekly issue.

Against the live catalog the report goes from **1 problem** to **All good — 50 services checked** with 11 inconclusive. The single 'problem' was this false positive.

Adds a regression test pinning the behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Referral links that redirect to a clean homepage are no longer incorrectly reported as dead.
  * These cases now appear as requiring manual verification while preserving the overall “All good” status.

* **Tests**
  * Added coverage to ensure collapsed referral URLs remain visible in reports without being counted as problems.

### DIFF
--- a/scripts/check_catalog_liveness.py
+++ b/scripts/check_catalog_liveness.py
@@ -91,9 +91,15 @@ def referral_code_lost(original: str, final: str) -> bool:
     """True when a referral link redirected to a bare homepage, dropping its code.
 
     The classic symptom of a retired referral programme: ``…/signup?ref=CODE`` ends
-    up at ``https://provider.com/``. Only that specific collapse is reported -- many
-    healthy links legitimately drop the query after setting a cookie, so anything
-    looser would be noise.
+    up at ``https://provider.com/``.
+
+    This detects the *collapse*, which is NOT the same as proving the code was
+    lost. A very common healthy pattern looks identical from outside: the site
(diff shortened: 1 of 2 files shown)


GeiserX:
Good catch — fixed. The footer contradicted the new classification.

Went slightly further than the suggestion: the legend is now a single shared constant rendered by **both** report shapes, and it is also emitted on an all-inconclusive run. Previously that path returned early, so a report with only inconclusive checks showed the 'Could not verify' table with no legend at all — exactly when a reader most needs to know that `unreachable` is not `broken`. Regression test added.

PR #134, opened 2026-07-31
fix(worker): do not use an ephemeral container ID as the worker identity

GeiserX:
## The bug

A worker keys its UI row — and its per-worker fleet key — on a `client_id`. When none was persisted yet, an already-enrolled worker fell back to `WORKER_NAME`, which defaults to `socket.gethostname()`. **Inside Docker that is the first 12 hex chars of the container ID, regenerated on every recreate.**

So every image bump minted a new identity. The worker presented its still-valid per-worker key under a `client_id` the UI had never enrolled, the UI correctly refused it, and every heartbeat returned `401 Unauthorized`.

**The failure is silent.** Service containers keep running and earning, so nothing looks broken — the fleet just loses the worker. Hit in production upgrading a live 3-worker fleet from 1.0.0 to 1.4.1:

```
POST http://cashpilot-ui:8080/api/workers/heartbeat "HTTP/1.1 401 Unauthorized"
```

The UI listed the worker as `159d39365879`; the recreated container had invented `515ccbc46cd9`.

## The fix

Narrow deliberately. The migration is worth keeping — on bare metal or a VM the hostname **is** stable, and reusing it preserves the row. Only the Docker container-ID shape is rejected, and only when `/.dockerenv` confirms we are in a container, so existing non-container workers migrate exactly as before (their test still passes unchanged).

Also: after three consecutive 401s while holding our own key, log the concrete remediation — which `client_id` we are sending and the file to write the expected one into — instead of repeating a generic warning forever.

## Verification

- 1317 tests pass; `ruff check` + `ruff format --check` clean
- New regression tests cover: container-ID rejected, real hostname still migrates, hex-shaped hostname **outside** Docker still migrates, persisted id always wins, first run persists for the next recreate
- Fix applied by hand to the live fleet first: all 3 workers recovered to `200 OK` with `key_confirmed=1`, and pre-seeding `/data/.worker_id` on the other two servers upgraded them with **zero** 401s

Docs: new *Worker identity* section in `docs/fleet.md` with the symptom and the recovery steps, and `CASHPILOT_WORKER_NAME` now flagged as one to set on Docker workers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worker identity handling in Docker environments to prevent unstable container names from causing authentication failures.
  * Added alerts after repeated authentication failures, including recovery guidance.
  * Successful heartbeats now clear authentication failure alerts.

* **Documentation**
  * Expanded worker configuration guidance for Docker deployments.
  * Documented persistent worker identities, authentication errors caused by identity changes, and recovery steps.

* **Tests**
  * Added coverage for stable identity persistence, migration scenarios, and ephemeral container-name detection.

### DIFF
--- a/app/worker_api.py
+++ b/app/worker_api.py
@@ -64,6 +64,11 @@ _ui_connected = False
 _last_heartbeat: str = "never"
 _last_error: str = ""
 
+# Consecutive 401s while holding our own per-worker key. One is unremarkable (the UI
+# may be restarting); a run of them means our identity no longer matches our key.
+_consecutive_auth_failures = 0
+_AUTH_FAILURE_ALARM_AFTER = 3
+
 # Per-worker fleet key. On first contact the UI enrolls this worker and hands back
 # a key unique to us, which we persist here (in our own private /data, never the
 # shared /fleet volume) and use for all subsequent auth in both directions. Until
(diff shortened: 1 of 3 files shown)


GeiserX:
All three addressed — two of them caught my own text being factually wrong:

1. **"It will enroll as a NEW worker"** — you're right, and production proved it. With a per-worker key present, `_active_key()` sends that key, the UI refuses it under an id it never enrolled, and the worker just 401-loops; no new row is ever created. That is exactly what I observed on watchtower. Reworded to say heartbeats will be **rejected until the id is restored**, and raised to `error` level.

2. **Counter reset** — fixed. Any non-401 outcome now resets it, in both the other `HTTPStatusError` path and the general exception path. Added sequence tests: `401,401,401` reaches the alarm; `401,timeout,401,500,401` stops at 1; `401,401,503,401` stops at 1; `401,401,200` resets to 0.

3. **Docs wording** — correct, that line described behaviour this PR removes. It now says what the variable actually does: keeps the display name stable across recreates, and gives a pre-existing worker a durable identity to migrate on. The recovery details stay in the Worker identity section.

1321 tests green, ruff clean.

PR #136, opened 2026-07-31
docs(agents): record what actually resisted the container hardening

GeiserX:
Follow-up to #135, written from actually doing it: hardened a live 40-container fleet (`cap_drop ALL` + `no-new-privileges` + `pids-limit`) in place across three servers. 40/40 hardened, 40/40 running, every mount identical before and after.

Two services resisted, and the distinction between them is the useful part:

**anyone-protocol** genuinely needs capabilities back. Its entrypoint chowns `/var/lib/anon` before dropping privileges:
```
chown: cannot read directory '/var/lib/anon': Permission denied
```
Minimum set is `CHOWN, DAC_OVERRIDE, FOWNER, SETUID, SETGID` — still far stricter than Docker's ~14 defaults.

**honeygain** looks identical from the outside (restart loop) but is **not** a capability problem at all — the provider still holds the previous session for that device name, and it resolves itself. Adding capabilities there is a wrong fix that then *sticks*, because a recreate tool preserving `CapAdd` carries them into every later run. Documented, along with the need for an escape hatch to drop them again.

Plus two mechanics worth knowing before recreating containers by hand: `rm -f` is SIGKILL (storage nodes log an unclean shutdown), and `Config.User` comes from the image so a non-root container stays non-root without passing `--user`.

Docs-only; sits above the generated beads blocks, still out of mkdocs nav.

## Summary by CodeRabbit

* **Documentation**
  * Added operational guidance for configuring service capabilities when dropping all Linux capabilities.
  * Documented troubleshooting steps for restart loops caused by stale server-side sessions.
  * Added container recreation cautions covering graceful shutdown, capability persistence, and preserving image-defined users.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -107,6 +107,48 @@ The durable fix is to make deployed storage and catalog agree — either redeplo
 service deliberately once, where an identity reset is acceptable, or keep the bind
 mount and never redeploy that slug from the catalog.
 
+### Applying the hardening: two services fight back
+
+Verified by hardening a live 40-container fleet in place.
+
+**anyone-protocol crash-loops under `cap_drop: ALL`.** Its entrypoint chowns
+`/var/lib/anon` before dropping privileges, which needs capabilities the blanket drop
+removes:
+
+```
+chown: cannot read directory '/var/lib/anon': Permission denied
+failed to change ownership of '/var/lib/anon' to anond:anond
+```
+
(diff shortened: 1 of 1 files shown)


GeiserX:
Closing the review threads on this PR: every finding is fixed on `main`, verified by reading the code there. Fixed in #137 — the MD040 fences were already given `text` identifiers; verified no bare opening fence remains in AGENTS.md.

PR #138, opened 2026-08-01
fix(packetstream): read the balance from the metric card, not a heading level

GeiserX:
PacketStream rebuilt its dashboard, so the collector could not find a balance and reported `Could not parse balance from dashboard — page structure may have changed`. Confirmed against the live dashboard (auth is fine — `/dashboard` returns 200; it was purely parsing).

**What changed on their side:** `window.userData` is gone entirely, and the balance moved into a metric card:

```html
<div class="card metric-card metric-card-balance"><h2 class=metric-title>Balance</h2>
<p class="card-subtitle mb-4 text-muted">Available Funds
<h2 class="default-font fw-600">$2.12</h2></div>
```

**The fix** anchors on the `metric-card-balance` class instead of a heading level. That matters: the heading has already moved once (`<h3>Balance</h3>` → `<h2 class=metric-title>`), which is what broke it, so the class is the more stable anchor. The markup is minified with unquoted attributes, so the pattern cannot require quotes, and the match is length-bounded so it cannot run past the card into the referral strip — which contains its own `$0.00` further down.

Older layouts stay as fallbacks.

**Verification:** extracts `2.12` from the real captured page, and a test puts a $99.99 referral amount *before* the balance card to prove the card still wins. 1330 tests pass, ruff clean.

### DIFF
--- a/app/collectors/packetstream.py
+++ b/app/collectors/packetstream.py
@@ -52,18 +52,36 @@ class PacketStreamCollector(BaseCollector):
             balance = 0.0
             parsed = False
 
-            # Pattern 1: server-rendered Balance card (current)
-            # <h3>Balance</h3>...<h2 ...>$0.13</h2>
+            # Pattern 1: the Balance metric card (current layout).
+            #   <div class="card metric-card metric-card-balance">
+            #     <h2 class=metric-title>Balance</h2>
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #140, opened 2026-08-01
docs: add a direction and roadmap page

GeiserX:
Adds `docs/roadmap.md`, published in the nav.

**Structure**

1. **What CashPilot is today** — two components, YAML catalog as source of truth, encrypted credentials, hardened containers. Plus an explicit *what it is not*: not hosted, not a wallet, and not a way to make meaningful money (a few euros a month per service; anyone promising otherwise is selling something).
2. **What we want it to become** — "the tool that makes running earning services on your own hardware safe, honest, and boring", with each word defined.
3. **Where we fall short** — deployment is stateless; nothing knows which state is irreplaceable; volume deletion is unguarded; collectors are fragile by nature.
4. **Seven avenues**, each with for / against / risk / verdict.
5. **How this improves things for users** — six concrete outcomes in the order a user notices them.
6. **Principles** — six tie-breakers for future decisions.

**The framing that matters:** the most serious class of bug here is not a missing key-management feature, it is that deployment has no *memory* — every redeploy rebuilds the container spec from the catalog, so a container whose real mounts differ from the catalog silently becomes a different container. Making deployment stateful removes the root cause; guarding the destructive paths is the cheap belt-and-braces.

**And what we will not build:** hosted storage of users' keys, any key recovery service, multi-account evasion tooling. Stated as a design boundary — the user holds everything, so there is no key we can lose because we never had it.

Docs-only. No personal infrastructure, credentials or third-party attributions.

## Summary by CodeRabbit

* **Documentation**
  * Added a Direction and Roadmap page outlining CashPilot’s architecture, security posture, product vision, known limitations, and prioritized improvements.
  * Documented planned enhancements for deployment state, destructive-operation safeguards, encrypted backups, catalog verification, resilient data collection, and expanded earning opportunities.
  * Clarified guiding principles and capabilities that are intentionally out of scope.
  * Added the new roadmap page to the documentation navigation.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,0 +1,284 @@
+# Direction and Roadmap
+
+This page describes what CashPilot is today, what we intend it to become, and the
+avenues we are weighing to get there — including the ones we have decided against
+and why. It is written to be argued with. If a trade-off here looks wrong, open an
+issue; the reasoning is laid out precisely so it can be challenged.
+
+## What CashPilot is today
+
+CashPilot is a **self-hosted dashboard for passive-income services**. You run it on
+your own hardware — a home server, a NAS, a VPS — and it does three things:
+
+1. **Deploys** Docker containers for services that pay you for idle resources:
+   bandwidth sharing, DePIN nodes, storage, and compute.
+2. **Collects** your earnings from those services' own APIs and dashboards, and
+   stores the history locally in SQLite.
+3. **Shows** it in one place, across every machine you run, with per-service and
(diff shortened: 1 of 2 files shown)


GeiserX:
Closing the review threads on this PR: every finding is fixed on `main`, verified by reading the code there. Fixed in #160 — record-vs-catalog precedence stated, hosted-custody legal claim qualified (FinCEN user vs administrator/exchanger, MiCA custody), export trust boundary settled, recovery boundary covers both secrets.

PR #142, opened 2026-08-01
fix(security): stop silently destroying stored credentials (CashPilot-1ii)

GeiserX:
## The problem

Three defects compound into silent, total loss of every stored credential — with a documented recovery path that does not exist.

**1. The docs describe a key the code never reads.** `README.md` and `SECURITY.md` both presented `CASHPILOT_SECRET_KEY` as the credential-encryption key. `app/database.py` read exactly one environment variable: `CASHPILOT_DATA_DIR`. `CASHPILOT_SECRET_KEY` is consumed by `app/auth.py:_resolve_secret_key` for **session signing** and lives at `/data/.secret_key` — a different key from the Fernet key at `/data/.fernet_key`.

A user who set a strong `CASHPILOT_SECRET_KEY` and backed it up believed their credentials were restorable. They were not: there was no supported way to supply or restore the encryption key at all. The decrypt-failure message even pointed them at that variable.

**2. An unwritable data directory silently yielded an ephemeral key.** The loader generated a key, failed to persist it, logged a `WARNING`, and returned the in-memory key anyway. Every restart minted a fresh key, so every stored `enc:` value became undecryptable. `decrypt_value` returns `""` on `InvalidToken`, so the surfaced symptom was a provider auth failure pointing nowhere near the cause.

This is unattended software. Nobody is reading a warning at boot.

**3. A corrupt key file was overwritten.** The old code fell through to "generate a new key", destroying the only artifact that could still decrypt credentials already in the database.

## The fix

- **`CASHPILOT_ENCRYPTION_KEY`** supplies or restores the key. An existing `/data/.fernet_key` **always wins** — that precedence is what makes the variable safe to introduce. Env-first would mean anyone setting it on a running instance instantly orphaned everything already encrypted; the restore case is unaffected because a wiped volume has no file to lose to. A divergence is logged, not silently resolved.
- **Startup refuses** when the key cannot be persisted (`verify_encryption_key_persisted()` in the lifespan). Deliberately *not* at import time — the suite imports `app.database` with the default `/data`, so importing must stay side-effect free. `CASHPILOT_ALLOW_EPHEMERAL_KEY` is the explicit opt-out, off by default.
- **A corrupt or unreadable key file is preserved** and startup refuses, rather than being overwritten.
- The key file is created `0o600` rather than chmod-ed after writing, which left it briefly readable depending on umask.
- A malformed supplied key is refused rather than quietly replaced.
- `decrypt_value` logs at `ERROR` and names the correct variable.

Docs (`README.md`, `SECURITY.md`, `CLAUDE.md`) now describe both keys, their files, and a backup/restore procedure. The new variable is wired into all three compose files.

## Note on the test suite

`tests/conftest.py` now points `CASHPILOT_DATA_DIR` at a writable temp dir. The suite previously passed *because of* the silent fallback being removed here: on a dev Mac `/data` is read-only and in CI it is not creatable, so the key was always ephemeral. Without this, the new guard correctly fails three lifespan tests.

## Upgrade note

If your data directory is not writable, CashPilot now **refuses to start** instead of quietly losing credentials on the next restart. Fix the mount, or set `CASHPILOT_ALLOW_EPHEMERAL_KEY=true`.

**Back up `/data/.fernet_key`** — without it, stored credentials cannot be decrypted.

## Verification

- `pytest tests/ --cov=app --cov-fail-under=90` — **1346 passed**, coverage 94.09%
- 11 new tests in `tests/test_encryption_key.py` covering: generation+persistence, restore via env key, file-wins precedence with divergence warning, refusal when unpersistable, the explicit opt-out, import staying safe, malformed key refusal, mismatch reporting, corrupt-file preservation, empty-file replacement, and `0o600` mode.

The corrupt-file and `0o600` findings came from reviewing this diff, not from the original report; both have regression tests.

## Summary by CodeRabbit

* **Security**
  * Improved credential encryption-key handling, including validation, secure file permissions, and preservation of invalid key files.
  * The application now refuses to start when encryption keys cannot be persisted, unless ephemeral-key use is explicitly enabled.
  * Added support for restoring encrypted credentials with a configured encryption key.

* **Documentation**
  * Clarified the distinction between session-signing and credential-encryption keys.
  * Added guidance for key backup, recovery, configuration, and secure deployment.

* **Tests**
  * Added coverage for key generation, restoration, validation, persistence, permissions, and failure scenarios.


GeiserX:
Addressed all four review findings.

**Major — stale malformed `CASHPILOT_ENCRYPTION_KEY` blocked startup.** Correct, and a real bug: the env value was validated before the file-precedence check and set the module-level error unconditionally, so a valid persisted key still refused to start on every boot. As you noted, that was also inconsistent with the valid-but-differing case, which only warns because the file wins. The error is now promoted to a startup failure only when there is no usable file key — i.e. when the supplied value was the one that actually mattered. When the file wins, the malformed value is logged as ignored. Regression test: `test_malformed_env_key_is_ignored_when_a_valid_file_key_exists`.

**Major — `CASHPILOT_ALLOW_EPHEMERAL_KEY` missing from Compose.** Correct; the escape hatch was documented but never reached the container, so it could not bypass the refusal it was meant to bypass. Added to all three files with a `false` default.

**Major — README restore command.** Correct; a bare assignment on its own line never reaches `docker compose`, so the restore would have silently minted a new key — the exact failure this PR exists to remove. Now on the same command line, with a note that exporting or `.env` works too.

**Minor — CLAUDE.md wording.** Correct; the env key is adopted whenever no key file exists, which covers a fresh install as well as a restore. Reworded.

Verification after the fixes: `ruff check . && ruff format --check .` clean, `pytest` **1347 passed**, coverage 94.10%.

GeiserX:
@coderabbitai review

PR #143, opened 2026-08-01
feat(security): refuse deletes that would destroy irreplaceable state (CashPilot-efx)

GeiserX:
Lands the first, independent slice of `CashPilot-efx`: the `delete_volumes` guard. The redeploy-side mount-diff guard is deliberately left for a follow-up, as the bead itself proposes.

## The problem

`remove` with `delete_volumes=true` (`app/orchestrator.py`) force-deleted every named volume with **no guard at all**. It is the only genuinely irreversible path in the codebase — the data behind those mounts has no server-side copy and no backup, so a mistake cannot be undone by redeploying.

What is actually at stake:

| Service | Mount | What is lost |
|---|---|---|
| Storj | `/app/identity` | Node identity. Proof-of-work bound (hours to regenerate) and carries the **held payout balance** |
| Storj | `/app/config` | Stored data and held amount; losing it disqualifies the node |
| Mysterium | `/var/lib/mysterium-node` | Node identity keystore — a new identity loses reputation and held earnings |
| Anyone Protocol | `/var/lib/anon` | Relay ed25519 identity; the relay's earning history restarts at zero |
| ProxyBase Markets | `/home/proxybase/.proxybase` | The generated wallet. There is no server-side copy — **this volume is the money** |

The browser UI happens never to send `delete_volumes`, but that is luck rather than design: the API accepts it from any client.

## The fix

- **Declared in catalog YAML, not hardcoded.** Services list `critical_volumes` with a `target` and a one-line `holds` description. Keyed by **mount target** because targets are stable while sources are not — Storj mounts `${IDENTITY_DIR}`.
- **Enforced in the worker.** Only the worker can see mounts, and a UI-side check would leave the other entry points unguarded.
- **Nothing is destroyed when it fires.** The check runs *before* the container is touched. An earlier ordering would have removed the container and only then refused.
- **409 with the reason**, listing each blocked volume and what it holds. `allow_delete_critical=true` proceeds deliberately.
- **Fails closed on unknown.** A slug the worker cannot resolve in its catalog is treated as *unsafe*, not *unprotected* — a trimmed image or a delisted service can never silently unlock an irreversible delete.
- The UI forwards this one structured worker error verbatim (other worker errors stay generic). A bare `409` with no reason would tell the operator nothing, which defeats the point.

## A pre-existing bug this exposed

Several tests point `catalog.SERVICES_DIR` at a tmp fixture directory and populate the module-level caches, which then persist for the **rest of the session**. Any later lookup of a real slug silently returned nothing.

That is worth calling out on its own: it means a catalog-keyed guard could stop guarding *without a single test failing*. `conftest.py` now drops the cache when it no longer matches `services/` — O(1), and only when genuinely polluted.

Two tests in `test_coverage_gaps.py` used the fictional slug `"test"` with `delete_volumes=True`; they now use a real non-critical service (`honeygain`). Their intent — `NotFound` counts as deleted, `APIError` is recorded — is unchanged; the fail-closed behaviour they were tripping over is covered directly in the new test file.

## Verification

- `pytest tests/ --cov=app --cov-fail-under=90` — **1346 passed**, coverage 93.73%
- 11 new tests: refusal, override, nothing-destroyed-on-refusal, non-critical unaffected, unknown-slug fail-closed, bind mounts skipped, removal without `delete_volumes` never blocked, plus a consistency test asserting every declared `target` matches a real mount — a typo there would otherwise protect nothing, silently.

## Not in this PR

The redeploy-side guard (refusing a redeploy that drops or retypes a critical mount, with a before/after diff) is the larger half of `efx` and lands separately, so this small irreversible-path fix is reviewable on its own.

## Summary by CodeRabbit

* **New Features**
  * Added protection against accidental deletion of critical container volumes.
  * Service removal now blocks deletion of protected data by default and reports affected volumes.
  * Authorized owners can explicitly override the protection when necessary.
  * Added critical-volume declarations for several services, including identities, wallets, and node data.

* **Documentation**
  * Documented critical-volume configuration and associated data-loss risks.

* **Bug Fixes**
  * Improved error handling and sanitized removal error responses.


GeiserX:
@coderabbitai review

GeiserX:
Closing the review threads on this PR: every finding is fixed on `main`, verified by reading the code there. Fixed in #154 — raw worker body no longer logged at warning (regression test added), schema overclaim corrected, noqa: E402 added, and the worker test client now restores lifespan_context in a finally.

PR #145, opened 2026-08-01
feat: persist the deployed spec and redeploy from it (CashPilot-tkd)

GeiserX:
> **Stacked on #143.** Base is `fix/critical-volume-guard`, not `main`, because this touches the same `main.py` deploy paths. GitHub will retarget it to `main` automatically once #143 merges. Review #143 first.

## The problem

Every redeploy rebuilt the container spec from catalog YAML, and **nothing recorded what was actually deployed**. When the running container had diverged from the catalog — a bind mount where the catalog declares a named volume, a host path that only existed because of an env substitution — the redeploy silently produced a *different* container. The worker destroys the old one before anything can compare them.

This is the root cause of the whole "lost node identity" class of bug. It was never a key-management problem: **CashPilot had no memory of what it deployed.**

The `deployments` table already carried an `env_vars_encrypted` column clearly intended for this. All three `save_deployment` callers omitted it, so it was never written.

## The fix

- **Record the full resolved spec at deploy time** in a new `spec_encrypted` column, Fernet-encrypted because env carries credentials. Added by automatic migration; `env_vars_encrypted` is superseded and left untouched.
- **Rebuild an existing service from that record**, with two deliberate exceptions: the **image** always comes from the catalog (or upgrades could never land), and **anything typed on this deploy wins** (or a rotated credential could never be corrected).
- **Disagreement is information.** Where the record and catalog differ, the difference is returned in the deploy response rather than resolved silently. The catalog is authoritative for what a *new* deployment looks like — not for this one.

## The subtle case I got wrong first

My first version had the record always win for mounts. That silently ignores an operator **deliberately relocating** their data. But the naive correction — "trust the catalog whenever env was supplied" — reintroduces the original bug the moment someone retypes a password.

So the check is keyed on the **volume-substituting env keys only**: supplying `IDENTITY_DIR` moves the data on purpose; supplying `PASSWORD` never costs the service its mounts. Both directions have tests.

## Degradation

A spec that cannot be serialised or decrypted falls back to today's behaviour — rebuild from the catalog — and logs loudly. It never fails the deploy, and never deploys half a remembered spec.

A caller that knows no spec (an external service, a status-only update) no longer blanks the stored one.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1368 passed**, coverage 93.75%
- 15 new tests covering the acceptance criteria: spec recorded and encrypted at rest (asserting the credential is not in the row), bind mount surviving a catalog that declares a named volume, ports reproduced, **Storj redeploying without retyping `IDENTITY_DIR`/`STORAGE_DIR`**, user input winning, new catalog keys still appearing, no-record and undecryptable-record both falling back, status updates not erasing the record, and the migration on a pre-existing database.

The deploy-route tests gain an autouse fixture defaulting to "never deployed", preserving exactly the behaviour they already assert.

## Summary by CodeRabbit

- **New Features**
  - Redeployments now preserve recorded volumes, ports, runtime settings, and environment values while applying intended updates.
  - Deployment specifications are saved securely and used to report differences between deployed and catalog configurations.
  - Added protection for critical volumes containing identities, keys, wallets, or other irreplaceable data.
  - Removal requests provide affected-volume details and require explicit confirmation to override safeguards.
  - Added critical-volume declarations for supported services.

- **Documentation**
  - Documented how to declare and protect critical container volumes.


GeiserX:
**Note on the base branch.** This started stacked on #143, but a non-`main` base means the CI workflows never trigger and CodeRabbit skips the review outright — so the work would have been verified only on my machine. Retargeted to `main` so it actually gets tested.

The consequence is that this PR now contains **3 commits**: the two from #143 plus this one. Review #143 first; once it merges, this diff reduces to the `tkd` commit alone. CI here therefore verifies the combination of both changes, which is what would actually be deployed.

GeiserX:
@coderabbitai review

GeiserX:
Both findings were correct and are fixed.

**Major — all-or-nothing volume reset.** Right, and a real bug. `moving_paths` was computed once for the entire `volumes` field, so supplying one path variable discarded the whole recorded dict. Storj is precisely the case that breaks: `IDENTITY_DIR` and `STORAGE_DIR` are bound to separate volumes, so moving the identity directory silently reset the storage mount to the unsubstituted `${STORAGE_DIR}` template — losing a mount the operator never mentioned, which is the exact class of silent divergence this feature exists to prevent. My rule was precise about *when* to relocate but not about *which mount*.

Each recorded mount is now matched by container target and kept unless the operator supplied a variable feeding *that* mount's host path. Mounts the catalog has added since deployment are still picked up, and the divergence message names the specific target. Regression test: `test_moving_one_path_does_not_reset_the_others`.

**Major — owner authorization for the critical-volume override.** Also right, and it was backwards: deploying already required owner, while permanently destroying a Storj identity or a generated wallet needed only writer. `_require_owner` is now checked whenever `allow_delete_critical` is set; an ordinary removal still only needs writer. Tests both ways. Committed on the #143 branch, since that is where the guard lives, and merged forward here.

Verification: ruff clean, **1372 passed**, coverage 93.61%.

PR #146, opened 2026-08-01
docs(security): state the security defaults and enforce them in tests (CashPilot-964)

GeiserX:
> **Stacked.** Contains the commits from #143 and #145 as well, because it builds on them and a non-`main` base gets no CI at all. Review those first; this diff reduces to the last commit once they merge.

Adds [docs/security-defaults.md](docs/security-defaults.md): three tiers — always on and not configurable, configurable but secure by default, and plain preference — under one principle:

> You may choose to weaken your own installation, but never by accident and never by default.

## The audit found something

The point of this bead was the audit, not the prose.

**Credential encryption at rest is decided by a naming convention.** `app/database.py` encrypts a config value when its key matches a known secret suffix — and *nothing enforced that convention*. A collector argument named `cookie`, `seed`, `mnemonic`, `jwt`, `passphrase`, `bearer`, `credential` or `private_key` would have been written to the database **in plaintext**, purely because of what it was called:

```
svc_api_key      encrypted=True
svc_password     encrypted=True
svc_cookie       encrypted=False   ←
svc_seed         encrypted=False   ←
svc_mnemonic     encrypted=False   ←
svc_private_key  encrypted=False   ←
```

That is worth fixing now rather than later: the v2.0.0 wallet work (`luj`, `e8u`) makes a field called `seed` entirely plausible.

- The suffix list is widened. **Reads are unaffected** — `decrypt_value` returns anything without the `enc:` prefix as-is, so existing rows keep working and only new writes change.
- `tests/test_secure_defaults.py` now **fails** if any collector credential would not be encrypted. The escape hatch is explicit: list the field in `PUBLIC_COLLECTOR_FIELDS` with the reason it is not a secret. Today that is `email`, `api_url` and `fingerprints` — identifiers and published data, not credentials.

## The rest of the posture, pinned in code

A stated posture nobody checks is just prose, so the same test file asserts: metrics off until explicitly enabled, alerting inert until a target is configured, an empty volume allowlist, loopback binding in **every** compose file, and no telemetry-shaped code anywhere in `app/`.

## A deviation I did not silently fix

The audit also found the project's own images referenced as `:latest` in all three compose files, contradicting both the tier-2 claim and the repo's own "semver tags, never `latest`" rule.

I filed it as **CashPilot-jz3** rather than changing it here: it is a user-facing change to the documented quickstart and it interacts with the release process, so it deserves a decision (pin and bump on release / floating major tag / drop the claim) rather than a unilateral patch. The docs page **names the gap** instead of overstating the guarantee.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1385 passed**, coverage 93.71%
- Page added to the mkdocs nav.

## Summary by CodeRabbit

* **New Features**
  * Deployments now preserve and reuse their recorded specifications, including ports, mounts, runtime settings, and environment values.
  * Added protection for volumes containing irreplaceable service data, with owner-only override support.
  * Expanded credential encryption coverage and secure-by-default protections.

* **Bug Fixes**
  * Fixed redeployment of services using encrypted deployment specifications.
  * Updated ProxyBase to its current client and image.

* **Documentation**
  * Added security-default guidance and documented protected service volumes.


GeiserX:
All eight findings fixed, across the three stacked branches. Two were on the tests themselves and both were valid — a test that passes for the wrong reason is worse than none.

**Major — hostname/privileged not preserved** (`tkd`). Correct on hostname: several bandwidth services key device identity to the container hostname, so a redeploy with an empty hostname field would silently give the service a new identity — the exact bug this feature exists to prevent. The recorded hostname is now kept when the operator did not type one, and a typed one still wins. `privileged` needs no handling: validation rejects it, so it is always `False`.

**Minor — non-hashable `error` crashes the sanitiser** (`efx`). Correct; `detail['error']` comes from a remote worker body and a list/dict made `not in` raise `TypeError`, turning a 409 into a 500. It now requires a string first. Test added.

**Minor — normalize mount paths** (`efx`). Correct, and it fails *open* on the only irreversible path: `/app/identity/` never matched `/app/identity`. Both sides are normalized (trailing slash stripped, root left alone). Test added.

**Major — metrics test never checked the unset default.** Fixed: it now asserts off-when-unset (the real fresh install), off for falsy values, and — so it is not vacuous — on when set.

**Major — Telegram test cleared the wrong variables.** Fixed: `notify.py` reads `CASHPILOT_TELEGRAM_BOT_TOKEN` + `_CHAT_ID`, not `_TOKEN`, so the check was never isolated.

**Minor ×3** — empty plaintext-exemption reasons now fail; `credential`/`keyfile`/`refresh_token` added to the suffix parametrization; and an integration assertion that a collector secret written through the config layer lands `enc:`-prefixed.

Verification: ruff clean, **1397 passed**, coverage 93.62%.

GeiserX:
Closing the review threads on this PR: every finding is fixed on `main`, verified by reading the code there. Fixed in #145 — hostname is preserved across a redeploy when the operator leaves it blank; privileged is always rejected so there is nothing to preserve.

PR #147, opened 2026-08-01
fix: clamp earnings per platform so a payout can't erase real earnings (CashPilot-glc)

GeiserX:
> **Independent of the deploy-stack PRs** (#142/#143/#145/#146). Branched off `main`, touches only the earnings math — reviewable and mergeable on its own.

## The bug

Daily and summary earnings were computed by summing every platform's balance delta and clamping the **total** at zero:

```
delta = max(0.0, current_total - previous_total)
```

The earnings table stores balance *snapshots*. When a payout lands, the paying platform's balance falls — and because the sum happens before the clamp, that drop cancels genuine earnings on other platforms:

```
platform A pays out:  20.00 → 0.10   (delta −19.90)
platform B earns:      5.00 → 7.00   (delta  +2.00)
old: max(0, −19.90 + 2.00) = max(0, −17.90) = 0.00   ← B's real $2 vanishes
```

So a day you actually earned could report **zero**, and one large payout hid everything else that day. The bead described this as "renders negative earnings"; the negative was already clamped, but the clamp on the *sum* is exactly what made it lossy.

## The fix

Clamp each platform's delta at zero **before** summing, in both `get_earnings_dashboard_summary` (today/month/yesterday) and `get_daily_earnings` (the chart series):

```
earned = sum(max(0.0, bal - previous.get(platform, 0.0)) for platform, bal in current.items())
```

A payout now counts as "nothing earned on that platform today", never a negative that eats the rest. A pure payout day reads as zero, never negative.

The summary queries do it in SQL (`SUM(MAX(delta, 0))`); the chart builds per-platform balances in Python rather than a cross-platform `SUM(balance) GROUP BY date`.

## Scope

This is the **correctness core** of `CashPilot-glc`. The larger ledger feature the bead also described — a payouts table, drop detection, a lifetime-earned vs current-balance split, and a projected payout date — is filed as **CashPilot-1og**, so this small, verifiable fix lands on its own.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1339 passed**, coverage 93.93%
- 4 new tests: B's earnings survive A's payout in both the summary and the chart, a pure payout day reads zero (and the whole series stays ≥ 0), and ordinary all-earning days are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to CashPilot are documented here.
 ## [Unreleased]
 
 ### Fixed
+- **A payout on one platform no longer erases earnings on another** (CashPilot-glc). Daily and summary earnings were computed by summing every platform's balance delta and only then clamping the total at zero. On a day a payout landed, that platform's balance fell, and the drop cancelled real earnings on other platforms before the clamp ever ran — so a day you actually earned could report zero. Each platform's delta is now clamped at zero *before* summing, so a payout counts as 'nothing earned there today' and never eats into the rest. A pure payout day reads as zero, never negative. The remaining half of the ledger work — a payouts table, drop detection, a lifetime-earned vs current-balance split, and a projected payout date — is tracked as a follow-on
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #148, opened 2026-08-01
feat: tell the user when a credential is about to expire (CashPilot-aug)

GeiserX:
> Independent of the deploy stack — branched off `main`, mergeable on its own.

## The problem

Several collectors need a value copied out of a browser, and some die within hours. Nothing told the user that.

A collector configured in the morning could be dead by evening. The only symptom was earnings quietly not being recorded — which looks **exactly** like a provider outage, even though the two need opposite actions:

- *your credential expired* → go re-copy a cookie
- *the provider redesigned their page* → wait for a fix, don't bother re-copying

The instance already hit in the wild: Bytelixir's session cookie expires ~2h after issue, while its `remember_web` cookie lasts about a year.

## The fix

- **`config.updated_at`** — config rows now record when they were last set, so a credential's *age* is knowable. Added by migration; existing rows get the migration time, not a fabricated older date, because that's the earliest we can honestly claim to know.
- **`CREDENTIAL_LIFETIMES`** — declares expected lifetime per service field, which are durable, and **why**, in words meant for the user rather than a maintainer.
- **`GET /api/credentials/health`** — turns age + declared lifetime into `fresh` / `expiring_soon` / `likely_expired` / `no_known_expiry`. It returns the key and the age only, **never the credential value**, so it is safe to drive a UI.
- Where a service offers a durable alternative and only the short-lived credential is configured, the report names it — and stops nagging once it's supplied.

`expiring_soon` fires at 75% of the declared lifetime, so the warning arrives *before* the collector goes silent. That's the entire point.

## Two tests guard the registry itself

- Every declared field must be a **real collector argument** — a typo would describe a credential nobody is ever asked for.
- Every declared field must carry a **reason** — a lifetime with no explanation cannot help anyone decide anything.

## Scope

This is the credential-honesty half of the bead. The alert-taxonomy half (distinguishing *expired-credential* from *provider-changed* in the alert text, and reporting which scrape pattern failed) touches the collector error paths and belongs with `CashPilot-bfl` (collector fixture/contract tests); it is not in this PR.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1352 passed**, coverage 93.94%
- 17 new tests: age recorded on single and bulk writes, values never exposed, migration on a pre-existing DB, the four status classifications, the durable-alternative suggestion appearing and then stopping, and unconfigured credentials not reported.


GeiserX:
@coderabbitai review

PR #149, opened 2026-08-01
feat: detect services that are running but no longer earning (CashPilot-kbs)

GeiserX:
> Independent of the deploy stack — branched off `main`.

## The gap

Catalog rot is the most common way a user silently stops earning, and the weekly liveness check cannot see the worst case: **a service that is up but no longer paying**.

The container runs. The collector authenticates. The balance never moves. Every other view of the system looks healthy, so nothing surfaces it — the user finds out by eventually noticing they stopped being paid.

## The detection

`get_flatlined_services()` reports a balance that has not moved across 7 recorded days, raised as a distinct `flatline` alert kind and exposed at `GET /api/earnings/flatlines`.

## The restraint matters as much

The bead is explicit that *a report which cries wolf is a report nobody reads*, so most of the logic is about what must **not** be reported:

| Case | Reported? | Why |
|---|:-:|---|
| Balance moved at all, even once | ❌ | Slow earners are still earning |
| Fewer than 7 recorded days | ❌ | A new deployment has not had time to earn |
| Balance always exactly zero | ❌ | Never paid — a setup problem, not a service that *stopped* |
| Only 3 readings across 30 days | ❌ | The window counts **distinct recorded days**, so a collection outage cannot masquerade as a flat balance |
| 7+ days at a non-zero, unchanging balance | ✅ | Running, authenticated, and not earning |

`record_alert`'s existing per-kind cooldown keeps this to one notification per service rather than one per collection cycle, and the whole check is wrapped so a diagnostic can never break the collection run itself.

## Scope

This is the **flatline** half of the bead. The gone-vs-down streak tracking lives in the weekly liveness script and needs state that survives between CI runs, so it is deliberately not in this change.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1342 passed**, coverage 93.83%
- 7 new tests, six of which assert the *non*-reporting cases above.

## Summary by CodeRabbit

- **New Features**
  - Added detection for services whose nonzero earnings remain unchanged for seven or more recorded days.
  - Added an authenticated API endpoint to view flatline earnings alerts.
  - Added notifications for detected flatlines, with cooldowns to prevent duplicate alerts.
  - Handles insufficient history, zero balances, outages, and database errors safely.

- **Documentation**
  - Documented flatline earnings detection and related release planning updates.


GeiserX:
@coderabbitai review

GeiserX:
Closing the review threads on this PR: every finding is fixed on `main`, verified by reading the code there. Fixed in #158 — the flatline query now requires a reading from today (a failing collector no longer reads as flat), and _flatline_check clears the cooldown for services that started earning again.

PR #150, opened 2026-08-01
docs: replace the competitor matrix with a comparison that stays true (CashPilot-qkc)

GeiserX:
## The problem, verified

The README table claimed money4band has **no web UI** and **no earnings dashboard**. I fetched its README rather than trusting the bead:

> "📊 Web Dashboard for easy monitoring."
> "Enjoy passive Earnings monitoring your performance through the web dashboard."

It also supports 20+ apps, not the 17 we listed. Both of our claims were false and **falsifiable in about thirty seconds** — which damages a reader's trust in everything else on the page.

## The fix

Replaced the matrix with prose comparing on what is *structurally* different — a fleet with central state rather than a single machine, earnings pulled from provider APIs rather than container liveness, breadth beyond bandwidth — and saying plainly that money4band is the more mature project and a good choice for one machine.

Deliberately **not** another feature matrix. A table asserting what other projects lack goes stale the moment someone ships a release, and nothing in CI can catch it. The section says so and invites an issue if it drifts.

## What I did NOT do, and why

The bead's second half says Grass, Teneo and Dawn should be re-verified because "that competitor removed them from its setup flow months ago."

**That premise does not survive checking** — all three are still in money4band's supported-app list today. So the stated evidence for demoting those three does not exist, and I left their catalog status alone rather than flipping it on a bad premise. Marking a service broken is user-facing and needs real evidence: a failing collector run, a dead endpoint, or a provider announcement. The catalog-liveness CI (`owv`, already shipped) is the mechanism that should produce that, as the bead itself notes.

Recorded on the bead so the correction isn't lost.

## Verification

- Referral links: **zero removed** (checked the diff explicitly — referral links are revenue), and the 11 referral-guard tests pass.
- `pytest --cov=app --cov-fail-under=90` — **1335 passed**, coverage 93.91%

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,18 +304,17 @@ Services that were evaluated but are no longer listed in the catalog due to bein
 
 ## How CashPilot Compares
 
-| Feature | CashPilot | money4band | CashFactory | income-generator | InternetIncome |
-|---------|:---------:|:----------:|:-----------:|:----------------:|:--------------:|
-| Web UI with guided setup | **Yes** | No (CLI) | Partial (links only) | No (CLI) | No (CLI) |
-| One-click container deploy | **Yes** | No (compose) | No (compose) | No | No (compose) |
-| Earnings dashboard | **Yes** | No | No | No | No |
-| Historical charts | **Yes** | No | No | No | No |
-| Multi-node fleet management | **Yes** | No | No | No | No |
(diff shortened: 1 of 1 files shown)


GeiserX:
@coderabbitai review

PR #151, opened 2026-08-01
docs(security): record what CashPilot will never do (CashPilot-kct)

GeiserX:
The roadmap already states three deliberate non-goals. This puts the reasoning where someone reads it *just before* proposing to cross one — `SECURITY.md`.

- **Never hold your keys on a server we control.** Software on your own machine, where key material never leaves it, is a fundamentally different thing from a service holding other people's secrets — the second is a regulated activity with licensing, capital and AML obligations. Staying on the self-hosted side is a constraint we design around, not an oversight.
- **Never offer key recovery.** Lose the key, lose the data. A recovery path we could operate would mean we could decrypt your data, which is the thing we just said we won't do. Saying so plainly beats implying a safety net that doesn't exist.
- **Never ship multi-account evasion tooling.** Several providers forbid more than one account per household; making evasion easy risks users' balances for a feature we can't make safe.

Plus an explicit contributor list of PR shapes that get declined with a link here.

## The one that will actually be tested

Once encrypted backup/restore ships (`qqo`), *"just let the server remember the passphrase so restore is easier"* is the obvious next request. It's called out by name, because convenience that requires us to be able to decrypt a user's data **is** the line.

## Verification

- `pytest` — **1335 passed** (docs-only change; no code touched)

## Summary by CodeRabbit

* **Documentation**
  * Added security principles clarifying that CashPilot will not hold server-side keys, recover keys, or provide multi-account evasion tools.
  * Added contributor guidance describing changes that will not be accepted.
  * Linked to the project roadmap for additional context.


GeiserX:
@coderabbitai review

PR #152, opened 2026-08-01
fix: pin the example compose files to major.minor, not :latest (CashPilot-jz3)

GeiserX:
Closes the deviation my own security-defaults audit turned up (#146 filed it rather than fixing it unilaterally, since it changes the documented quickstart).

## The problem

The shipped compose files used `drumsergio/cashpilot:latest`, contradicting **both** the project's own rule (*semver tags, never `latest`*) and the pinning claim in the security posture. A user following the quickstart:

- couldn't tell what they were running
- could take a **breaking change on a routine `docker compose pull`**

## The choice

Pinned to the **major.minor** tag, `:1.4`:

| Option | Patch fixes | Surprise minor/major | Goes stale |
|---|:-:|:-:|:-:|
| `:latest` | auto | **yes** | no |
| `:1.4` (chosen) | auto | no | at next minor |
| `:1.4.4` exact | manual | no | at next patch |

`major.minor` keeps the property a self-hosted user actually wants — security patches arrive without intervention — while making a minor or major upgrade a deliberate edit.

**Verified before shipping:** `docker manifest inspect` confirms both `drumsergio/cashpilot:1.4` and `drumsergio/cashpilot-worker:1.4` exist and are multi-arch. Pointing the quickstart at a tag that didn't exist would have been worse than the original bug. The tag is already published by `build.yml` alongside `:latest` and the exact version.

## Guardrail

`tests/test_compose_image_pins.py` fails if `:latest` reappears, if a CashPilot image has no explicit tag (untagged *is* `:latest`), or if the UI and worker drift onto different tags — a version-skewed pair is a support problem nobody wants.

## Honest remaining gap

Keeping the pin current at each **minor** release is still a manual step. The test makes drift visible rather than silent, but it can't bump the tag for you. Automating that in the release workflow means CI committing back to `main`, which I didn't want to add unasked.

## Upgrade note

Existing deployments are unaffected until you edit your compose file. To keep tracking every release as before, set the tag back to `:latest` deliberately.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1341 passed**, coverage 93.91%
- Both compose files parse as YAML; 6 new pin tests


GeiserX:
@coderabbitai review

PR #153, opened 2026-08-01
feat: pre-deploy reality check (CashPilot-w58)

GeiserX:
## The problem

The catalog shows one generic earnings range per service. A user can't tell **before** deploying whether it will work in *their* situation — and several won't. They find out weeks later, when it has earned nothing.

`GET /api/services/{slug}/preflight` answers "what will this realistically do for me?" from what we already know: the catalog `requirements` block, and what's already running on the same worker.

| Situation | Verdict |
|---|---|
| Same service already here, `devices_per_ip: 1` | **will earn nothing** — and some providers forfeit the balance for this |
| Same service already here, no declared limit | reduced earnings — one connection, shared |
| Storage node | disk commitment named, plus: part of the balance is held back and **forfeited if abandoned early** |
| GPU / residential-IP requirement | named as a precondition CashPilot cannot verify |
| Nothing notable | looks fine |

## Two rules the module is built around

**Informed consent, not a nanny.** Even the worst verdict doesn't block — it says *"this will most likely earn nothing here. You can deploy it anyway."* `blocking` is always `False`, and a test asserts it.

**Never imply a check that wasn't run.** Egress IP type and connection speed can't be detected yet (that's `5qc`), so requirements depending on them are reported as *unverified preconditions in the user's own words*, and every response carries a `not_checked` list. A clean result must never read as a guarantee about things nobody looked at.

Absent catalog data deliberately **softens** the verdict rather than inventing one.

## A real gap this surfaced

Most services **don't declare `devices_per_ip`** — only 3 of 50 — even though the README table lists it as 1 for Honeygain, IPRoyal, EarnApp and others. Verified directly:

```
honeygain requirements -> {residential_ip, devices_per_account: 10, min_bandwidth: '', gpu, min_storage}
```

So the strongest verdict is unavailable for exactly the services where a user is most likely to make the mistake. Filed separately rather than bulk-transcribed from the README — that table is hand-maintained and is the same class of artefact that turned out to be factually wrong in #150. Each value should be confirmed against the provider, and left absent where it can't be.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1352 passed**, coverage 93.68%
- 17 new tests, including every catalog service assessing without error (a malformed requirements block must not break the deploy page)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ All notable changes to CashPilot are documented here.
 - Prometheus HTTP metrics bound their `path` label: `/api/workers/{id}` collapses per-id paths and unknown/scanner paths fold into a single `/{other}` label, so probe traffic can't grow label cardinality without limit
 
 ### Added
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #154, opened 2026-08-02
fix(security): stop logging raw worker error bodies, and 3 review follow-ups

GeiserX:
CodeRabbit's review of #143 landed **while that PR was merging** and I did not wait for it — my mistake. All four findings were valid. This addresses them on `main`.

## Major — the sanitisation was undone one line above it

`_safe_worker_detail` deliberately withholds unstructured worker error bodies from the caller because they *"can carry internal paths and hostnames"*. The line right above then wrote the full `resp.text` to the **warning log** for every worker error:

```python
logger.warning("worker proxy error (%s): %s", resp.status_code, resp.text)
```

A deploy failure echoes env values, which for several providers **are live credentials**. So the careful sanitisation protected the HTTP response and leaked the same data into the log.

Now: warning logs the status and the *sanitised* detail; the raw body is debug-level. A regression test asserts a credential-shaped value in a worker error body never reaches the warning log.

## The other three

- **`services/_schema.yml` overclaimed.** It said `critical_volumes` also makes the worker refuse a redeploy that would drop the mount. It doesn't — only the delete guard shipped. Corrected, with the gap stated plainly instead of left as a promise the code doesn't keep.
- **Imports after env setup** in the test are intentional (the API key must be set before `worker_api` imports) — now marked `noqa: E402` rather than relying on ruff's current defaults.
- **The worker test client leaked state.** It replaced `worker_api.app.router.lifespan_context` and never restored it. `app` is module-level and shared for the whole session, so that leaked into every later test building a client from it. Now a fixture that restores the original in a `finally`.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1370 passed**, coverage 94.11%

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1210,10 +1210,16 @@ async def _proxy_to_worker(
             else:
                 resp = await client.post(f"{url}{path}", json=json, params=params, headers=headers)
             if resp.status_code >= 400:
-                logger.warning("worker proxy error (%s): %s", resp.status_code, resp.text)
+                # NOT resp.text at warning level: withholding the raw body from the
+                # caller (see _safe_worker_detail) is pointless if it goes straight
+                # into the log instead. A deploy failure can echo env values, which
+                # for several providers are live credentials.
+                safe = _safe_worker_detail(resp)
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #161, opened 2026-08-02
feat: report net profit, not just gross earnings (CashPilot-f5u)

GeiserX:
Every dashboard in this space reports what a service **paid**. None report what it **cost to run**.

For a large share of users the honest number is negative. A service earning **EUR 2/month** on hardware drawing **15W** at **EUR 0.30/kWh** costs EUR 3.29 in electricity — a **EUR 1.29 monthly loss** that we were presenting as income.

`GET /api/earnings/net` now reports gross, estimated cost and net per service, and names the ones whose trailing net is below zero.

## Three honesty rules, tested directly

The failure mode here is dishonesty rather than a crash, so these are asserted, not assumed:

- **An estimate is labelled as one.** The CPU-share model is crude — directionally right, and far better than pretending cost is zero — so every figure carries `cost_quality` and a UI cannot render it with false precision.
- **Net is reported alongside gross, never instead of it.**
- **No tariff configured → cost is `unknown`, not zero.** A zero would render gross as net and overstate earnings, which is the exact thing this bead exists to stop.

## Two modelling decisions worth stating

- **Idle draw is shared, not billed per container.** A machine kept running for ten containers costs its idle draw once, not ten times.
- **An unmetered host is charged nothing.** A VPS bill does not move with CPU, so charging estimated watts against it would invent a cost the user is not paying — as dishonest as ignoring a real one.

## Scope

This is the calculation and the API. Surfacing it in the dashboard belongs with the `app.js` work (`cyc`). Real measurement from RAPL or a smart plug can replace the estimate later **without changing the contract**, since `cost_quality` already distinguishes `estimated` from `measured`.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1515 passed**, coverage 93.70%
- 19 new tests, including the bead's motivating case end-to-end and every honesty rule above.

## Summary by CodeRabbit

- **New Features**
  - Added net-earnings reporting through a new authenticated API endpoint.
  - Reports gross earnings, estimated electricity costs, net earnings by service, currency, reporting period, and loss indicators.
  - Distinguishes measured and estimated costs, identifies unknown tariffs, excludes VPS electricity costs, and allocates shared idle power across containers.

- **Documentation**
  - Added an unreleased changelog entry describing net-profit reporting.


GeiserX:
@coderabbitai review

GeiserX:
Holding this PR rather than merging it. Five findings, and two of them mean the numbers it reports would be wrong — in a change whose whole purpose is reporting numbers honestly.

**`power.py:153` — per-service net is gross when the tariff is unknown.** Correct, and it is my own stated rule broken in my own code. I got the totals right (`total_net: None`, `cost_known: false`) but each service row still reports `net == gross`. That is precisely the 'render gross as profit' failure this bead exists to stop. Per-service `cost` and `net` must be `None` when there is no tariff.

**`main.py:1835` — window-scoped cost subtracted from a lifetime balance.** Also correct and more damaging. `get_earnings_per_service()` returns each platform's latest *balance*, not what it *earned* over `days`. So a 30-day electricity cost is subtracted from a lifetime (or current) balance, and the resulting 'net' means nothing. The currency mixing in the same path compounds it.

The other three are real too: stopped containers inflate `container_count` and understate cost; collapsing every worker into one count and one host TDP charges a 3-node fleet a single idle floor and makes `is_metered` per-worker impossible; and there is no error boundary around worker-status data.

Fixing these properly needs a window-scoped earnings query and per-worker power attribution — more than a patch on top of this diff. Not merging a feature that would mislead about money.

GeiserX:
Reworked. Four of the five findings are fixed and verified; the fifth is called out below rather than quietly left.

**Fixed — per-service net was gross when the tariff is unknown.** This was the feature's own rule broken in its own code: totals said `cost_known: false` / `total_net: null`, but every row still reported `net == gross`. Rows now return `cost: null`, `net: null`, `cost_quality: "unknown"`, and nothing is flagged negative on an unknown cost. Regression test asserts `out["services"][0]` directly.

**Fixed — a window's cost charged against a lifetime balance.** `get_earnings_per_service()` returns the latest *balance*, a running total. New `get_earned_by_platform(days)` sums per-platform deltas across the window, clamped per platform the same way the dashboard does (`glc`), so a payout does not read as negative earnings. Two tests: a 100→103 climb reports 3 rather than 103, and a payout-then-earn sequence reports the earning, not a negative.

**Fixed — stopped containers inflated `container_count`**, shrinking each running service's share of the idle floor and understating cost. Now filtered on `status == running`.

**Fixed — no error boundary.** A worker-status failure no longer takes out the earnings figures, which come from the database and are still reportable.

**Still open, deliberately: per-worker power attribution.** Every worker is still collapsed into one count and one host TDP, so a multi-host fleet is charged a single idle floor and `is_metered` cannot be applied per worker. That needs the worker grouping carried through to the watt calculation — a larger change than this, and I would rather it be its own reviewable diff than a rushed addendum here.

Verification: ruff clean, **1520 passed**, coverage 93.67%.

PR #163, opened 2026-08-02
fix(ci): publish both images on every release, and fail if a tag is missing

GeiserX:
Releasing v1.7.0 exposed this, and it breaks the **documented quickstart** for anyone following the README.

## What happened

The build only rebuilds an image whose source changed — which is right. Nothing under `Dockerfile.worker` or `app/{worker_api,orchestrator,constants,catalog,fleet_key}.py` had changed, so `build-worker` was correctly skipped. The result:

```
drumsergio/cashpilot:1.7.0         EXISTS
drumsergio/cashpilot-worker:1.7.0  MISSING
```

But the compose files pin **both** images to the same major.minor tag (`jz3`), and `tests/test_compose_image_pins.py` **enforces** that they match. So on any release where one image is rebuilt and the other isn't, following our own documented pinning yields a compose file referencing a tag that doesn't exist — and `docker compose pull` fails.

The build contract and the pinning rule contradicted each other, and the test enforced the broken side.

## The fix

**`retag-ui` / `retag-worker`** re-point an unchanged image at the new release's tags via `buildx imagetools create`. That's a **manifest operation, not a rebuild** — an unchanged multi-arch artefact is never rebuilt merely to satisfy a naming convention, which would trade minutes of QEMU for nothing.

**`verify-tags`** then asserts every tag the release claims actually resolves, and fails the run otherwise.

## Why the gate matters more than the fix

CI passed happily while v1.7.0 was half-published. It was found only by deploying and running `docker manifest inspect` by hand. That is exactly what a release gate should catch, and now does.

This is the third bug this session that CI could not see and running the thing could.

## Verification

- full pytest green; the 6 compose-pin tests still pass, and with retagging their same-tag rule becomes genuinely valid rather than a trap
- workflow parses; jobs: `resolve-tags, lint, build-ui, retag-ui, retag-worker, build-worker, verify-tags`

## Summary by CodeRabbit

* **Chores**
  * Improved release image tagging so unchanged UI and worker images remain available across all release tags.
  * Added automated validation to confirm that every expected UI and worker release tag is present.
  * Release builds now provide more consistent image availability and fail early when tags are missing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,48 @@ jobs:
           cache-from: type=gha,scope=ui
           cache-to: type=gha,mode=max,scope=ui
 
+  # A release must publish BOTH images even when only one changed. The compose
+  # files pin them to the same tag, so a version that exists for one and not the
+  # other makes the documented quickstart reference a tag that is not there —
+  # `docker compose pull` then fails for anyone following the README.
+  #
+  # This is a manifest re-point, not a rebuild: imagetools copies the existing
+  # multi-arch descriptor to the new tags in seconds, so an unchanged image is
+  # never rebuilt just to satisfy a naming convention.
+  retag-ui:
+    needs: [lint, resolve-tags]
+    if: ${{ !inputs.build_ui && inputs.version != '' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4
+      - uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Re-point the unchanged UI image at this release's tags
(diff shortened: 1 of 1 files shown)


GeiserX:
Fixed, and it turned up a second bug of mine.

**Your finding was right.** `ui_tags`/`worker_tags` derive from `inputs.version`, which `workflow_dispatch` lets a caller set freely, and all three new blocks interpolated them straight into `for tag in ...` — in steps that run *after* the registry login. They now come through `env:` and are read as data with `while IFS= read -r`.

**Fixing it exposed something worse in my own `verify-tags` block.** My first attempt piped into the `while`, which runs it in a subshell, so `missing=1` never reached the check — the gate would have passed silently on a missing tag, which is the exact failure it exists to catch. And the `[ -s file ] && missing=1` I reached for would have **failed the step under `bash -e` on the happy path**, when nothing is missing and the file does not exist. A gate that fails when everything is fine is worse than no gate at all.

Now written to a file and read with a redirect: no subshell, no conditional-as-last-command. Simulated locally against real Docker Hub tags — `cashpilot:latest`, `cashpilot:1.7.0`, `cashpilot-worker:1.6.0` all resolve and the flag stays 0.

PR #164, opened 2026-08-02
feat: producer state — is it earning, not just running (CashPilot-b4e)

GeiserX:
`get_health_scores` is computed from restarts, crashes and check results — so a container that has produced **nothing for a month still scores 100/100**. "It runs but earns nothing" is the most common complaint in this category, and container health is structurally incapable of seeing it.

This is a deliberately **separate** verdict. A service can be perfectly healthy as a container and dead as a producer; collapsing the two hides exactly the case the user cares about.

## Two of the bead's three signals

- **Earnings movement** over a trailing window, for services that have a collector.
- **`health_signals` in the service YAML** — log regexes with a plain-language `means` and a state. This is the **only** signal available for the 30+ catalogued services with no collector, and keeping the patterns in YAML follows the rule that service-specific knowledge never lives in `app/`.

## What it refuses to claim

A service with **no collector reads `unknown`, never `idle`**. Telling someone their service is idle when we simply cannot see its earnings is the same false confidence this exists to remove. Too little history is `unknown` too, and a stopped container isn't judged at all.

Every verdict carries its reasons — a state with no reason isn't actionable. And a concrete log diagnosis outranks an earnings observation, so *"login failed"* beats *"earnings moved"*: the log says **why**.

## No catalog entries yet, on purpose

I have not populated `health_signals` for any service. A guessed regex that never matches is dead weight; one that matches the wrong line tells the user something false. Populating them needs real log samples per service. The schema says so explicitly, and a test fails if any declared pattern doesn't compile or has no `means`.

## Scope

The third signal — container network rx/tx counters — needs those figures wired through the Docker worker heartbeat, which the bead itself calls "the bulk of the work". Not in this change.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1556 passed**, coverage 94.38%
- 27 new tests, with the endpoint covered up front (four earlier PRs failed `codecov/patch` for exactly the omission of endpoint tests)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -39,6 +39,7 @@ from app import (
     notify,
     power,
     preflight,
+    producer_state,
     setup_token,
 )
 from app.worker_proxy import _pin_url_to_ip, _validate_worker_url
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #165, opened 2026-08-02
feat: per-service disclosure — what it does with your machine (CashPilot-66x)

GeiserX:
*"Is this malware?"* is the most common reaction to software in this space, and it's fair: the user is asked to run **closed-source third-party containers that route traffic through their home connection**. We answered with an earnings range and a setup guide.

Five questions per service, in the catalog YAML like everything else service-specific: what is actually sold, **whether strangers route traffic through the user's IP**, what identifying data leaves the machine, what it means for their ISP contract, and the account rules.

## The design point is what happens when the answer is missing

- An **undocumented service** reports `documented: false`, names every unanswered question, and says plainly that this is **not a statement that it is safe**. A service nobody has documented is exactly the one to be careful with.
- A **partly filled** block admits the gaps rather than looking complete.
- `routes_third_party_traffic` is **three-valued**: "not documented" must never collapse into "no". That's the dangerous default on the single most consequential fact in this catalog.
- An explicit *"unknown — the provider does not say"* is kept as a **real answer**. Somebody looking and failing is different from nobody looking.

## Four services documented, 46 deliberately not

Documented because their mechanism is unambiguous from the provider's own description or from work verified directly:

| Service | Resells your IP? |
|---|---|
| mysterium | **Yes** — you are a dVPN exit node; traffic is attributed to your line |
| anyone-protocol | **Yes** — relays others' traffic; exits attract abuse complaints |
| proxybase-xyz | **Yes** — buyers proxy through your connection; that is the product |
| storj | **No** — you store encrypted fragments; nobody routes through you |

The other 46 are **left undocumented rather than guessed**. `GET /api/disclosure/coverage` reports 4 of 50 and enumerates the gap — presenting the documented subset as if it were the whole catalog would imply a completeness that doesn't exist.

## Verification

- `pytest --cov=app --cov-fail-under=90` — **1578 passed**, coverage 94.44%
- 22 new tests, endpoints covered up front. A test also fails on any unknown disclosure key, since a typo'd field would be silently dropped rather than shown.

### DIFF
--- a/app/catalog.py
+++ b/app/catalog.py
@@ -51,6 +51,12 @@ def _validate(data: dict[str, Any], path: Path) -> list[str]:
     if status is not None and status not in _VALID_STATUSES:
         errors.append(f"{path.name}: invalid status {status!r} (expected one of {sorted(_VALID_STATUSES)})")
 
+    disclosure = data.get("disclosure")
+    if disclosure is not None and not isinstance(disclosure, dict):
+        # `disclosure: TODO` loaded fine and then 500'd the endpoint at request
+        # time, so the fault was invisible until someone opened that one service.
+        errors.append(f"{path.name}: disclosure must be a mapping, not {type(disclosure).__name__}")
+
     docker = data.get("docker")
     if isinstance(docker, dict):
         image = docker.get("image")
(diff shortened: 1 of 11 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #166, opened 2026-08-02
feat: see the fleet the way providers do — by IP, not by machine (CashPilot-5qc)

GeiserX:
Every bandwidth provider caps **per IP address, not per device**. Honeygain treats a second active device on a network as *"network overused"*; EarnApp documents that extra devices behind one IP share a single daily cap.

CashPilot's whole fleet model encourages deploying a service to several machines, and warned about none of it. Two workers in one house are two rows on the dashboard and **one customer** to the provider — so the second earns nothing.

Seeing this requires knowing about the *other* machines, which is why a single-host tool structurally cannot do it.

## What lands

- Each worker reports its public egress IP and, where it can tell locally, whether it is a hosted machine.
- Deploying to a worker that shares an address with one already running the service is called out **by name**, before the deploy.
- `GET /api/fleet/egress-groups` groups the fleet by exit rather than by host.
- A residential-only service on a machine that identifies itself as a VPS is no longer a polite "check this yourself" — it's a verdict.

## Three rules that keep the warning from being wrong

A warning that cries wolf gets switched off, so:

- **An undetected IP is neither shared nor distinct.** Those workers produce *no finding at all* and are bucketed as undetermined.
- **A private address is a detection failure, not an identity.** Concretely: every worker in the reference fleet has a `100.64/10` tailnet address — grouping on it would collapse the entire fleet into one fabricated conflict.
- **An absent `devices_per_ip` means undocumented, not unlimited.** `0` is a verified no-limit; a missing key is not. Only 4 of 50 services declare one.

Detection is local-first: the hosting hint comes from DMI vendor strings, so nothing is disclosed to a third party and it works offline. A bare hypervisor (QEMU, VMware) is deliberately **not** hosting — a VM on a home server is a residential connection.

## Two defects fixed, one confirmed live in 1.7.0

- **`/api/services/{slug}/preflight?worker_id=N` returned 500.** Confirmed against the deployed UI on real worker rows: `AttributeError: 'str' object has no attribute 'get'`. `list_workers` returns JSON *text* columns; the endpoint treated them as dicts.
- **`/api/services/{slug}/producer-state` never matched a container** — it filtered on `c["service"]` while heartbeats emit `slug`, so every service reported UNKNOWN. Merged after 1.7.0, so it never reached production.

Both had passing tests, because both fixtures hand-fed a shape production never produces. Tests now build rows the way SQLite actually returns them.

## Review

An independent review raised 22 findings; each was reproduced by execution before being acted on. The serious ones: a stalled IP lookup could stall the **serial** heartbeat loop and take a worker offline (httpx's timeout is per-operation, not a deadline), and `not_checked` claimed the connection type was checked while a finding beside it said it couldn't be. Also fixed: an off-by-one that under-warned, Android workers invisible to the check, a custom IP endpoint silently falling back to third parties, and **a unit test making a real network request**.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1704 passed**, coverage 94.60%
- Verified against real hardware: both home servers report `ASUSTeK COMPUTER INC.` and correctly classify as *not* hosting; the DMI path is readable from inside the worker container; the worker image module set was simulated to prove `app/egress.py` adds no import the worker lacks.

Detection is on by default and makes one outbound call per hour, so it is documented in the README (not just `CLAUDE.md`) with `CASHPILOT_EGRESS_DETECT=off` and two ways to avoid the third party entirely.

## Summary by CodeRabbit

- **New Features**
  - Added fleet-wide egress IP grouping and a new endpoint for viewing shared, unknown, and worker-specific egress details.
  - Worker heartbeats now report egress IP and network type.
  - Added configuration for network classification, IP detection, explicit egress addresses, and custom lookup endpoints.
- **Bug Fixes**
  - Preflight checks now detect shared egress conflicts, per-IP limits, duplicate instances, and hosted-network mismatches.
  - Improved worker and service-state handling, including legacy container data compatibility.
  - Heartbeats recover from temporary detection or network failures.
- **Documentation**
  - Documented egress tracking, privacy controls, fleet grouping, limits, and configuration options.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -171,8 +171,8 @@ Three invariants that must not be "simplified" away:
   the tailnet range every reference-fleet worker uses; grouping on it would put
   the whole fleet into one fabricated conflict.
 - **An absent `devices_per_ip` is undocumented, not unlimited.** `0` is a
-  verified no-limit and suppresses the warning; only write it when the provider
-  actually says so.
+  verified no-limit and *downgrades* the warning to a shared-bandwidth note (it
+  does not silence it); only write it when the provider actually says so.
 
 Note that heartbeat container entries key the service as **`slug`**; code that
 filtered on `service` silently matched nothing in production. Use
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -149,6 +149,34 @@ Two separate Docker images:
 | `CASHPILOT_UI_URL` | Yes | -- | URL of the CashPilot UI |
 | `CASHPILOT_API_KEY` | Yes | -- | Shared API key for worker-to-UI auth |
 | `CASHPILOT_WORKER_NAME` | No | hostname | Human-readable name for this worker |
+| `CASHPILOT_WORKER_NETWORK` | No | detected | `residential`/`hosting`; overrides the DMI-vendor guess |
+| `CASHPILOT_EGRESS_DETECT` | No | on | `off` disables the hourly public-IP lookup |
+| `CASHPILOT_EGRESS_IP` | No | -- | State the public IP directly (validated; a LAN/tailnet address is rejected) |
+| `CASHPILOT_EGRESS_IP_URL` | No | -- | Custom IP-echo endpoint returning a bare IP |
+
+### Egress IP awareness (CashPilot-5qc)
+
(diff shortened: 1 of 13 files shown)


GeiserX:
@coderabbitai review

PR #167, opened 2026-08-02
fix: record only per-IP limits a provider actually states (CashPilot-4qv)

GeiserX:
The deploy preflight could not give its strongest verdict **exactly where the mistake is most likely**, because the limit it needs was recorded for 4 of 50 services. Two things were wrong: values were missing, and one that was present had no source.

## Added — each traced to a first-party statement

Quoted in full in `docs/research/per-ip-device-limits.md`.

| Service | `devices_per_ip` | Source |
|---|---|---|
| honeygain | **1** | Help centre: *"Users may connect 1 device per one IP/Network."* |
| iproyal (Pawns) | **1** | ToS: *"limits the number of devices per single IP address to one (1)."* |
| earnfm | **1** | Help centre: *"One Device Per Network."* |
| ebesucher | **1** | Blog: each device needs *"a unique IP address."* |

## Removed — `repocket: devices_per_ip: 2`

**It has no provider source.** Repocket documents a five-device limit per *account*, which is a different thing; the "2 per IP" figure traces only to review sites repeating each other. This is the exact failure the bead exists to prevent, and it was already shipping — a wrong number tells the user something false with full confidence, while an absent one only degrades the verdict to "check the provider's terms".

Same reasoning leaves `proxyrack`, `packetstream`, `bitping` and `urnetwork` absent, each with a note recording what *is* documented. **A test asserts these stay absent**, so the next person to find the number on a blog has to find a source first.

Where a provider documents *sharing* without a number (`earnapp`, `spide`), the key stays absent and the effect goes in the note — "your second device earns nothing extra" is not a device count.

## Separately, and more seriously: EarnApp forbids what this project does

Verified first-party against EarnApp's own help centre:

> **No. Installing EarnApp on Virtual Machines (VMs), Docker containers, or hosting services is strictly prohibited.**
>
> Prohibited: VMs · **Docker containers** · Cloud hosting · **Personal or home servers** · Any device used for **business or monetization** purposes.
>
> Penalty: account **terminated without prior notice**, pending payments **canceled**.

CashPilot deploys every service as a Docker container on a home server. Shipping EarnApp silently would make this tool the cause of the ban.

It is recorded as `requirements.container_prohibited` and reported as the strongest verdict the preflight has. **It still does not block** — this is a risk a user is entitled to accept knowingly, but not one they should discover afterwards.

The flag requires a first-party source and must never be inferred from a residential-IP rule, so a test fails if any service picks it up without one.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1626 passed**, coverage 94.45%

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Services CashPilot can deploy and manage automatically via Docker.
 | [Anyone Protocol](https://anyone.io) | [Guide](docs/guides/anyone-protocol.md) | ✅ | ✅ | Unlimited | 1 | Crypto (ANYONE) |
 | [Bitping](https://app.bitping.com) | [Guide](docs/guides/bitping.md) | ✅ | ✅ | Unlimited | 1 | Crypto (SOL) |
 | [Earn.fm](https://earn.fm/ref/GEISYB91) | [Guide](docs/guides/earnfm.md) | ✅ | ✅ | Unlimited | 1 | Crypto |
-| [EarnApp](https://earnapp.com/i/TSMD9wSm) | [Guide](docs/guides/earnapp.md) | ✅ | ❌ | 15 | 1 | PayPal, Gift Cards, Wise |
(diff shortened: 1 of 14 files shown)


GeiserX:
@coderabbitai review

PR #168, opened 2026-08-02
feat: is anything actually crossing the wire? (CashPilot-t6y)

GeiserX:
The third producer-state signal, and the only one needing **neither a collector nor a hand-written log pattern** — every Docker container has network counters. That matters because 30+ catalogued services have no collector at all, so for them this is the only evidence available.

Docker's counters are **cumulative** totals since container start, so one reading says nothing and two say something only if nothing reset in between.

## Three rules, each from a way this signal lies

**A counter that went backwards is a restart, not negative traffic.** Subtracting the old baseline gives a large negative number; silently clamping it to zero would call a busy service idle. Backwards means the baseline is gone → `UNKNOWN`.

**Missing counters are not zero counters.** Docker omits the `networks` key *entirely* for host-network containers. On the reference fleet the single busiest service — a dVPN exit moving **~15 MB/s** — is host-networked and reports nothing. Reading absence as no-traffic would mark the best earner dead.

**The idle threshold is measured, not assumed.** The bead asked for this explicitly, so I sampled every managed container on a live host over two minutes:

| Container | rx B/s | tx B/s |
|---|---:|---:|
| mysterium | 815,589 | 15,275,715 |
| honeygain | 6,709 | 6,773 |
| earnfm | 546 | 661 |
| earnapp | 19 | 12 |
| proxybase | 5.5 | 7.3 |
| **traffmonetizer** | **0.0** | **0.0** |
| **iproyal** | **0.0** | **0.0** |

**Eleven of fourteen moved measurable traffic while earning little or nothing** — keepalives and control chatter. A threshold of zero would have called all of them active. Two moved *literally nothing*. That's the only distinction the data supports, so the threshold sits just above it and the signal answers exactly one question: did anything at all cross the wire?

Full table and method in `docs/research/idle-network-traffic.md`.

## What it is allowed to conclude

Deliberately narrow. Bandwidth resale is buyer-driven, so a healthy node moves nothing while nobody is buying:

- `SILENT` supports **IDLE** and never FAILING, and never overrides observed earnings or a failure log.
- `MOVING` never asserts PRODUCING — **bytes on the wire are not money**. For a collector-less service it says only that the container is demonstrably not inert, which beats a bare "unknown".

The baseline cache is in memory on purpose: losing it on restart costs one unknown reading, whereas persisting it risks pairing a stale baseline with counters that reset while we were down.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1771 passed**, coverage 94.72%
- 36 new tests, endpoints covered up front — including a host-networked container proving it does not claim silence, and a restart-mid-interval proving it does not clamp to zero.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -14,6 +14,7 @@ import logging
 import os
 import re
 import secrets
+import time
 from contextlib import asynccontextmanager
 from datetime import UTC, datetime, timedelta
 from time import monotonic
@@ -38,6 +39,7 @@ from app import (
     exchange_rates,
     fleet_key,
     metrics,
+    net_activity,
     notify,
     power,
     preflight,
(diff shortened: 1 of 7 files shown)


GeiserX:
@coderabbitai review

PR #169, opened 2026-08-02
feat: notice when a provider changes its API, and test credentials on demand (CashPilot-bfl)

GeiserX:
The collector tests assert subclassing, async-ness and attribute presence. **None of that notices a provider renaming a field** — which is exactly how the ProxyBase/Repocket/EarnFM breakages shipped: the collector stopped finding the balance and every test stayed green.

Three parts, and it's worth being blunt about what each can actually prove.

## 1. Contract fixtures — what they can prove

`tests/fixtures/collectors/*.json` pin the **field names** each collector depends on, plus its currency and a representative shape. Tests assert:

- every collector has one (a new collector can't ship without a contract);
- no fixture describes a collector that no longer exists;
- every declared field still appears in the collector's own source, so a refactor that stops reading one fails CI;
- the recorded sample actually contains the fields it claims — a fixture documenting an unparseable shape is *worse* than none, because it looks like evidence.

**What they cannot do is notice a provider changed its API today.** They're frozen at the moment they were written. That's said plainly in the module docstring rather than left implied.

## 2. The nightly live check — the part that touches reality

Runs the real collectors behind a `live` pytest marker, **excluded by default** so a PR never depends on a third party being up, and opens (or comments on) a `collector-drift` issue when a shape changes.

On `ubuntu-latest` deliberately: free on a public repo, ephemeral, and it keeps provider credentials off the self-hosted fleet. **The issue body does not paste the log** — a failing collector can echo a request payload.

## 3. Credential self-test

`POST /api/services/{slug}/test-credentials` answers *"are these valid?"* in a second, instead of waiting up to an hour for the scheduler to say so via a notification bell.

Two constraints shaped it more than the feature did:

- **It cannot echo the credential or the provider's raw body.** A failed login frequently replies with the submitted payload, and a raw body has leaked into logs from a worker path in this repo before. Outcomes are *classified*; the original text is logged at debug only; the response has no field able to carry a secret. A test asserts a token embedded in an error never reaches the caller.
- **A test button invites a retry loop**, and retrying a rejected login repeatedly is how providers flag an account. Rate-limited per service, with a test asserting the provider is hit **once**, not twice.

The collector is built *uncached* via a new `collectors.build_one`, reusing the same argument table as `make_collectors` so the two can't disagree about which keys a service needs. Cached would have validated the credentials the user just replaced.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1818 passed**, coverage 94.37%
- Endpoint tests up front, including the leak and rate-limit cases.

## Summary by CodeRabbit

- **New Features**
  - Added an authenticated option to test service credentials on demand.
  - Results now show clear statuses for valid, invalid, missing, unsupported, or unreachable services.
  - Successful checks can display the current balance and currency.
  - Credential test attempts are rate-limited to prevent repeated requests.

- **Bug Fixes**
  - Sensitive credentials and raw provider errors are excluded from responses.

- **Chores**
  - Added scheduled live checks to detect collector issues automatically.

### DIFF
--- a/.github/workflows/collector-live-check.yml
+++ b/.github/workflows/collector-live-check.yml
@@ -1,0 +1,83 @@
+name: Collector Live Check
+
+# A recorded fixture is frozen at the moment it was written, so it can prove our
+# code still reads what we said it reads and nothing more. Only calling the real
+# API notices that a provider renamed a field — which is how several collectors
+# silently started returning zero while every test stayed green.
+#
+# Runs on a schedule and by hand, never on a PR: it needs real credentials, and
+# a third party being down must never redden someone's pull request.
+
+on:
+  schedule:
+    - cron: "17 5 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: collector-live-check
+  cancel-in-progress: false
+
+jobs:
+  live:
+    # Public repo -> GitHub-hosted runners are free, ephemeral and isolated,
+    # which also keeps provider credentials off the self-hosted fleet.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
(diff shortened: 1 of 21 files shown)


GeiserX:
@coderabbitai review

PR #170, opened 2026-08-02
feat: encrypted export of irreplaceable service state (CashPilot-qqo)

GeiserX:
Some services keep state that **exists nowhere else** — a node identity, a relay key, a wallet generated inside the container on first run. Lose the disk and the held payout balance and the node's reputation go with it. Almost nobody backs these up, because nothing tells them the files exist.

It's also the most dangerous thing in this codebase: it hands the worker a genuinely new capability — reading key material out of a container volume. The worker already holds the Docker socket, so the question was never whether it *could*, but whether the design gives anyone a reason to.

## The constraints, none of them tunable

**The installation cannot decrypt its own exports.** No server-held wrapping key, no fallback to the Fernet key protecting credentials. Otherwise compromising CashPilot yields every node identity it ever exported — making the backup feature the highest-value target in the system. The user supplies the target: an X25519 **recipient public key** (no secret enters CashPilot at all) or a **passphrase**, used and discarded.

**The only egress is the authenticated response.** No upload, sync or webhook — *absent*, not disabled. A test walks the route bodies and fails on any call that could send a bundle somewhere, so the guarantee is checkable rather than a policy nobody audits.

**No scheduled export.** A timer needs a stored secret — precisely the server-held key the first rule forbids.

**Encryption happens on the worker**, so the UI handles ciphertext only and plaintext key material never crosses the fleet network.

**AEAD with the header authenticated**, so a bundle can't be silently relabelled to claim it belongs to a different service. ChaCha20-Poly1305 + scrypt, both already in the existing `cryptography` dependency — zero new deps, no musl build risk.

## Reading the state safely

The container is **paused** while its state is read, and unpaused even when the read raises. A keystore copied mid-write yields a backup that opens perfectly and restores a *corrupt* node — a failure that surfaces only when you finally need it. Pause rather than stop, so a node doesn't lose uptime score for a backup.

Only mounts the catalog marks `critical_volumes` are read — never the whole container — and that's the same inventory the delete guard refuses on, so "irreplaceable" has one definition rather than two that drift apart.

## Verify, because a file is not a backup

"I have a backup file" and "I have a backup that works" are different claims, and people discover the difference on the worst possible day. Verify decrypts **in memory on the worker**, compares digests only, writes nothing, and distinguishes:

| Answer | Meaning |
|---|---|
| matches | restores exactly what is running now |
| does not match | opens fine but is **out of date** — not an error |
| cannot open | wrong passphrase/key, or the file was altered |

## Not included, deliberately

**Restore.** Writing into a volume is the destructive half and deserves its own review rather than riding along with the read path.

## Note on the tests

Several security properties are asserted against the **AST**, not the source text — a substring search for `Fernet` or `upload` also matches the docstrings *explaining* why neither exists, so it fails on the very prose documenting the guarantee.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1776 passed**, coverage 94.42%
- Worker image module set simulated to prove `state_backup.py` adds no import the worker lacks.

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -43,6 +43,9 @@ COPY --chown=cashpilot:root app/fleet_key.py ./app/
 # Egress identity: pure helpers, deliberately free of app-level imports so
 # the worker image stays minimal.
 COPY --chown=cashpilot:root app/egress.py ./app/
+# Encrypted state export: encryption happens on the WORKER so the UI only
+# ever handles ciphertext.
+COPY --chown=cashpilot:root app/state_backup.py ./app/
 COPY --chown=cashpilot:root app/worker_api.py ./app/
 COPY --chown=cashpilot:root app/orchestrator.py ./app/
 COPY --chown=cashpilot:root app/catalog.py ./app/
--- a/app/orchestrator.py
+++ b/app/orchestrator.py
@@ -7,6 +7,7 @@ inspection for cashpilot-managed containers via the Docker SDK.
 
 from __future__ import annotations
 
+import contextlib
 import logging
 import os
 import time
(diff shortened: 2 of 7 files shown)


GeiserX:
@coderabbitai review

PR #171, opened 2026-08-02
feat: payouts, lifetime-vs-balance, and how far off the cashout is (CashPilot-1og)

GeiserX:
The earnings table stores balance **snapshots**, so a payout is invisible — the balance simply drops. Two things follow, and both mislead quietly.

**One number was doing two jobs.** "Earned" went *down* when the user got paid, which is the opposite of what the word means. Lifetime earned and current balance answer different questions, and neither is answerable while they're conflated.

**Nobody could see how far away the payout was.** A 20 USD minimum can be two or three months on a single device, and not knowing that is what makes people quit before they ever cash out.

## Nothing is ever auto-confirmed

A balance also falls for a provider correction, a reversed fraud check, or a session that resets a counter. An unconfirmed guess written as income **corrupts lifetime-earned permanently and invisibly**.

So a drop is recorded as `probable`, surfaced in the notification bell, and waits for a human. A rejected one is *deleted* rather than flagged, so no later query can count it. Only one unconfirmed payout is kept per platform — otherwise every collection cycle files another copy of the same event and the user learns to dismiss the prompts.

Detection also requires a **documented** threshold (the provider's own `cashout.min_amount`). Guessing one — "any fall over 20%" — would fire on ordinary noise.

## It refuses to project rather than project badly

A confident wrong date is worse than silence, because the user plans around it. These are three different answers with three different fixes, not three ways of writing zero:

| State | Meaning |
|---|---|
| `not_enough_data` | too little history to say anything yet |
| `not_earning` | it will not get there at the current rate |
| `too_far` | over a year away — worth knowing before leaving it running |
| `projected` | "about 34 days, at 0.5 a day" |
| `reached` | cash out now |

The trailing rate **skips negative steps**: a cashout inside the window would otherwise read as weeks of negative earnings.

## What the catalog taught this change

`earncc` declares `min_amount: 0`. That's a documented *"no minimum, cash out anything"*, not a missing value — so `min_payout` is three-valued like every other threshold here. Collapsing zero into `None` would tell a user with no minimum that there's nothing to count down to, when in fact they can cash out right now. A zero still can't drive *detection*, since every drop would qualify.

A test asserts every catalogued `min_amount` parses, because a malformed one silently disables both detection and the projection.

## Details

- Payout values store the **fx rate at the time they landed**, mirroring the earnings table, so a token payout isn't restated every time the token moves.
- Detection failing can never break earnings collection.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1772 passed**, coverage 93.84%

## Summary by CodeRabbit

* **New Features**
  * Added payout detection when earnings balances decrease.
  * Added alerts for probable payouts without automatically counting them as income.
  * Added options to review, confirm, or reject detected payouts.
  * Added payout history, confirmed earnings totals, balance history, and cashout progress projections.
  * Added support for currencies and historical exchange rates.

* **Bug Fixes**
  * Prevented payout-detection issues from interrupting earnings collection.
  * Excluded payouts and balance decreases from daily earning-rate calculations.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -355,6 +355,26 @@ CREATE TABLE IF NOT EXISTS user_preferences (
 -- earnings flatlines later). Persisted rather than kept in memory so they survive a
 -- restart: passive income is unattended, and an alert that only exists in a running
 -- process is an alert nobody ever sees.
+CREATE TABLE IF NOT EXISTS payouts (
+    id           INTEGER PRIMARY KEY AUTOINCREMENT,
+    platform     TEXT    NOT NULL,
+    amount       REAL    NOT NULL,
+    currency     TEXT    NOT NULL DEFAULT 'USD',
+    -- USD per 1 unit of `currency` WHEN THE PAYOUT LANDED, mirroring the
+    -- earnings table. A MYST payout valued at today's rate would silently
+    -- restate history every time the token moves.
+    fx_rate_usd  REAL,
+    -- 0 until a human says this really was a payout. A balance also falls for
+    -- provider corrections and reset sessions, and recording a guess as income
+    -- corrupts lifetime-earned in a way the user cannot see.
+    confirmed    INTEGER NOT NULL DEFAULT 0,
+    method       TEXT    NOT NULL DEFAULT '',
+    detected_at  TEXT    NOT NULL DEFAULT (datetime('now')),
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #172, opened 2026-08-02
feat: generate the README service tables from the catalog (CashPilot-9q1)

GeiserX:
*"YAML is the single source of truth"* is the rule this project is built on — and the README service tables were the one place it was violated. Hand-maintained, so they drifted. Not hypothetically: **the README kept publishing a per-IP device limit for weeks after the catalog dropped it for being unsourced.** Same wrong number, more visible place.

Adding a service also meant remembering four places. Every extra place is one a first-time contributor gets wrong, picks up a review comment for, and doesn't come back from.

## What lands

- Tables generated **between markers**, so surrounding prose and footnotes survive.
- A **CI check** that runs the strict catalog loader, proves the tables match, and fails if a service has no guide — separate from the full test job so a contributor gets an answer in seconds.
- A **guide stub generator** driven by the same YAML.
- A **services PR template** stating plainly that an unsourced value must be *left out*, not guessed.

Generating it surfaced **31 rows asserting a per-IP limit of `1` that no catalog entry backs.** Those now render `?`, with a footnote clarifying `?` is *not* a synonym for "no limit".

## Three things the generator got wrong first

All caught before shipping, and each is the same class of bug this project keeps hitting:

- **An absent flag rendered as a definite "no."** Now three-valued: ✅ / ❌ / `?`.
- **A semantic mismatch.** The README's "Residential IP ✅" meant *"residential works"*; the YAML's `residential_ip` means *"REQUIRES residential"* — different questions, with opposite answers for a third of the catalog. Columns renamed to say what the data means, and VPS applies the default the schema **documents** (opposite of `residential_ip`) rather than guessing or shrugging.
- **EarnApp's footnote marker was deleted by the first regeneration** — the marker pointing at the warning that it forbids Docker containers, VMs and home servers, which is exactly how CashPilot deploys it. Markers are **derived from the catalog** now, so the most consequential sentence in the file can't be regenerated away. A test asserts it survives.

## Revenue check

These tables carry every signup link, so:

- **Zero referral URLs lost.**
- **One gained** — the README linked a bare `https://proxybase.xyz` while the catalog had the referral URL. Generation fixed a live leak.
- Tests assert every active service's referral URL appears, and that none is dropped relative to the committed README.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1758 passed**, coverage 94.64%

## Summary by CodeRabbit

* **New Features**
  * Added catalog-driven README tables for Docker, browser/desktop, and GPU services.
  * Added generated service guide templates with requirements, earnings, and service details.
* **Documentation**
  * Improved catalog transparency with source information, status indicators, and explanations for unknown values.
  * Added a checklist for submitting new services.
* **Bug Fixes**
  * Added automated checks to detect outdated catalog tables, missing guides, and invalid service entries.
* **Tests**
  * Added coverage for table synchronization, formatting, links, warnings, and guide generation.


GeiserX:
@coderabbitai review

PR #173, opened 2026-08-02
feat: is this machine worth keeping powered on? (CashPilot-l01)

GeiserX:
The question people ask is *"does this service make money"*, and on a shared box **that question has no honest answer**. Adding a bandwidth container to a machine that is already on costs roughly 1-3 W — about €0.19/month at €0.105/kWh — which is **below the measurement error of a consumer smart plug**. A per-service net figure there would look authoritative and be noise.

So this refuses to produce one, and **the refusal is the feature**. `per_service_is_meaningful()` only returns true above the resolution of the cheapest instrument that could verify the number. GPU compute would clear that bar, but no compute service in this catalog is Docker-deployable today — so that half stays theoretical rather than shipping as if it worked.

## The question that does have an answer

A dedicated ~65 W node at €0.20/kWh costs about **€9.50/month** against a typical $3-6 gross from a single node. Net negative, knowable, and nobody is currently told.

## Distinctions it keeps

Collapsing any of these produces a confident wrong number:

- **Dedicated vs shared.** A machine that exists only for these services can be turned off, and the summary says so. A NAS would be on anyway — its draw is the price of the box, not of the services. Blaming the services would push someone into a pointless decision.
- **Unknown vs zero.** No tariff or no wattage yields `unknown`, *names which one is missing*, and says guessing would be worse than silence. In the fleet roll-up, machines with unknown cost are **counted separately rather than folded in as zero**, so the total never quietly understates what the fleet costs.
- **Metered vs not.** A VPS bill doesn't move with CPU, so charging it estimated watts invents a cost the user isn't paying.
- **Confident vs indistinguishable.** Within 25% of break-even the answer is *"the estimate cannot really tell these apart"*, not a decimal pretending otherwise.

## It never acts on the number

No auto-stop, no throttling. A test parses the module and fails if it ever calls `stop`/`remove`/`restart`/`kill`/`pause`. The electricity is the operator's and so is the decision.

## Not included

Reading watts from a **Home Assistant** power sensor. It's the right upgrade path — HA already solves tariffs and PVPC, and reimplementing that would be foolish — but it needs a URL/token config surface that should be designed rather than guessed at here.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1769 passed**, coverage 94.69%

### DIFF
--- a/app/machine_economics.py
+++ b/app/machine_economics.py
@@ -1,0 +1,206 @@
+"""Is this MACHINE worth keeping powered on? (CashPilot-l01)
+
+The question people ask is "does this service make money", and for a shared box
+that question has no honest answer. Adding a bandwidth container to a machine
+that is already on costs roughly 1-3 W — about EUR 0.19 a month at EUR
+0.105/kWh — which is **below the measurement error of a consumer smart plug**.
+A per-service "net profit" figure there would be fabricated precision: it would
+look authoritative and be noise.
+
+The question that does have an answer is about the machine. A dedicated ~65 W
+node at EUR 0.20/kWh costs about EUR 9.50 a month against a typical USD 3-6
+gross from a single node. That is net negative, it is knowable, and nobody is
+currently told.
+
+So this module answers per-MACHINE and refuses to answer per-service on a shared
+host. That refusal is the feature, not a limitation of it.
+
+Three rules:
+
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #174, opened 2026-08-02
feat: own the attribution risk, and contain the lateral one (CashPilot-q0o)

GeiserX:
The documented danger of running proxyware **is not container escape**. It's that someone else's traffic exits the user's address and is attributed to them — the FBI/IC3 has a public notice on residential proxy networks, and researchers have shown operators cannot see what leaves their own exit node.

The second real risk is a hostile image scanning the LAN it was handed: Home Assistant, NAS shares, a media server, a password manager.

## The first risk can't be engineered away — so say it

Selling that egress **is the product**. Any control that blocked it would block the reason the user installed this. So CashPilot states it plainly at the deploy step rather than in an FAQ. Owning the downside is what persuades a self-hosting audience.

The notice is driven by the service's **own disclosure entry**, which gives three outcomes rather than two:

| Case | Result |
|---|---|
| documented as reselling your IP | full warning, including that the suspended ISP account is *yours* |
| documented as **not** reselling (storage node) | **silence, deliberately** — a false warning teaches people to click past real ones |
| **undocumented** | a notice saying exactly that |

That third case matters most: most of the catalog is undocumented, and letting *"nobody checked"* render as *"no risk"* would be the worst possible reading.

## The second risk is what can actually be reduced

The module computes what can be confined **without touching anything**. Creating bridges and firewall rules on someone's host has real blast radius, and missing an exception costs money rather than convenience — a storage node that can't accept inbound connections stops earning and puts its held balance at risk.

So it names every exception and why, and leaves the act to the operator. Tests parse the module and fail if it ever calls `run`/`Popen`/`create`/`connect`, or imports `docker` or `subprocess`.

Three verdicts, each true of the real catalog:

- **honeygain** — outbound only, isolatable as-is
- **storj** — needs its inbound port *and* its local dashboard reachable (that's where CashPilot reads earnings)
- **mysterium** — declares `network_mode: "host"`, so no container bridge applies; the honest answer is a VLAN or a separate machine

## The thing that's easy to get wrong

**A Docker bridge alone does NOT stop LAN access.** The rule that matters is on the host firewall. The blocked list includes `169.254.0.0/16` because a container reaching cloud metadata on a VPS can often mint credentials for the whole account.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1761 passed**, coverage 94.71%

### DIFF
--- a/app/lan_isolation.py
+++ b/app/lan_isolation.py
@@ -1,0 +1,218 @@
+"""Lateral containment, and owning the attribution risk (CashPilot-q0o).
+
+The documented danger of running proxyware is not container escape. It is that
+someone else's traffic exits YOUR address and is attributed to you — the FBI/IC3
+has issued a public notice about residential proxy networks, and researchers
+have shown operators cannot see what traffic leaves their own exit node. The
+second realistic risk is a hostile image scanning the home LAN it was handed:
+Home Assistant, NAS shares, a media server, a password manager.
+
+The first risk cannot be engineered away here. Selling that egress IS the
+product, so any control that blocked it would block the thing the user installed
+CashPilot to do. What can be reduced is LATERAL reach: a bandwidth container has
+no business talking to a NAS.
+
+Two decisions shape this module.
+
+**It states the attribution risk plainly, at deploy time.** Not in an FAQ, not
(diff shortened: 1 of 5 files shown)


GeiserX:
@coderabbitai review

PR #175, opened 2026-08-02
feat: optional container runtime, explicitly unsupported (CashPilot-54q)

GeiserX:
The research verdict this implements is **mostly a refusal**: do not adopt gVisor as a default or as a supported "hardened profile".

## Why not

- **The escape path is already closed.** Deploys refuse `privileged`, drop every capability and add back only what a service's catalog entry declares, allowlist `network_mode`, and block host bind mounts including the Docker socket. There is no documented case of a mainstream proxyware image escaping its container.
- **It doesn't address what actually happens.** The evidenced risks are IP attribution and lateral movement. gVisor addresses neither — `q0o` does.
- **~1.7x network throughput cost** on a workload that is *pure network I/O*: you'd pay for it in the exact dimension you're paid in.
- **It breaks things.** `runsc` only *simulates* `NET_ADMIN`, and host networking negates the isolation anyway — so a host-networked service either fails or gains nothing.
- **It doesn't install cleanly where this runs:** Unraid has no package manager and isn't systemd; a Raspberry Pi needs a custom 48-bit-VA kernel; on macOS Docker Desktop already runs a Linux VM.

## What ships instead

The cheap concession: an optional `runtime` on the deploy spec, so an advanced user who has **already** installed one can opt a single service in and own the outcome.

- **The allowlist is read from the daemon**, not hardcoded — and a test asserts the validator carries no runtime name. A hardcoded list would accept a runtime the host never installed, failing later with a Docker error nobody can act on.
- **Nothing defaults to it.** An absent runtime doesn't even query the daemon; `deploy_raw` passes `None` so Docker uses its own default — what every service is tested against.
- The endpoint listing what a host provides reports `supported: false` and says plainly this is **not a hardening recommendation**.
- The runtime is validated *before* anything else, and a test proves it can't be used to slip past the existing guards.

The docs state the reasoning, not just the switch — the useful thing here is knowing why the obvious hardening step is the wrong one.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1749 passed**, coverage 94.67%

### DIFF
--- a/app/orchestrator.py
+++ b/app/orchestrator.py
@@ -155,6 +155,23 @@ def _read_pids_limit() -> int:
 _PIDS_LIMIT = _read_pids_limit()
 
 
+def available_runtimes() -> set[str]:
+    """Runtimes the local Docker daemon actually reports.
+
+    The allowlist is derived from the daemon rather than hardcoded, because a
+    runtime CashPilot has heard of but the host has not installed would fail at
+    container-create time with a Docker error the user cannot act on. Asking the
+    daemon means the only runtimes offered are ones that exist here.
+    """
+    try:
+        info = _get_client().info()
+    except Exception:
+        logger.warning("Could not read Docker runtimes; treating none as available")
+        return set()
+    runtimes = info.get("Runtimes")
+    return set(runtimes) if isinstance(runtimes, dict) else set()
+
+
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #176, opened 2026-08-02
feat: drop unsafe-inline from the CSP (CashPilot-guw)

GeiserX:
`script-src 'unsafe-inline'` means **any markup an attacker gets injected executes** — and this UI renders provider-supplied strings.

Removing it needed two things at once, because a nonce covers script *elements* and never inline event *attributes*:

- **53 inline handlers gone**, migrated to one delegated listener keyed on `data-action` — plus four async-font `onload=` attributes and a form `onsubmit=` that the first pass missed entirely.
- **The five remaining inline `<script>` blocks carry a per-response nonce**, read from `request.state` so the header value and the markup value can never disagree. If they did, every inline script would be blocked and the UI would silently stop working.

`style-src` keeps `'unsafe-inline'` **and says why**: the templates carry inline `style=` attributes, which a nonce cannot cover, and a style injection cannot execute. An exception left unexplained becomes permanent.

## The verification came first

This bead sat blocked for several sessions on *"cannot verify without a browser"* — which was true and mattered. So `scripts/ui_check.sh` drives **headless Chrome over CDP** and reports what a browser actually sees: inline handlers present, CSP violations, uncaught errors. No dependencies (Chrome is already installed; Node has a global `WebSocket`). It exits non-zero on a violation, so it **gates** a change rather than describing one.

It earned itself twice:

1. It caught an inline handler still firing after the `onclick` migration looked complete — the font-preload `onload=` and the form `onsubmit=`. Neither would ever be found by searching for "onclick".
2. **It caught the real bug.** Putting the delegated listener in `app.js` left the onboarding and login pages without it — those templates are standalone and never load `app.js`, so their buttons rendered perfectly and **did nothing at all**. No error, no visual difference, nothing a static check or a code review would see. The listener now lives in `delegate.js`, loaded by every entry template, and tolerates `CP` being undefined.

## Handler resolution

CP first, then a page-local global — a few controls belong to one template and have no business on the shared namespace. Anything else **logs an error rather than doing nothing**: a silently dead button is the failure this refactor could most easily have introduced fifty times over.

## Verified, not inferred

- Zero inline handlers, zero CSP violations, zero JS errors on every reachable page.
- A delegated button proven to still work end to end: the wizard advances **step-1 → step-2**.

Static guards keep it that way: no template or rendered string may carry an inline event attribute, no inline script may lack a nonce, every entry template must load the listener, and every `data-action` must name a handler that exists.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1767 passed**, coverage 94.65%

### DIFF
--- a/app/deps.py
+++ b/app/deps.py
@@ -39,6 +39,20 @@ _TRUST_PROXY = os.getenv("CASHPILOT_TRUSTED_PROXY", "").strip().lower() in ("1",
 templates = Jinja2Templates(directory="app/templates")
 
 
+def _csp_nonce(request) -> str:
+    """The per-response CSP nonce, for stamping inline <script> blocks.
+
+    Read from request.state so the value in the markup always matches the one in
+    the header. If they ever diverged every inline script would be blocked and
+    the UI would silently stop working, which is exactly the failure a template
+    author would not think to test for.
+    """
+    return getattr(getattr(request, "state", None), "csp_nonce", "")
+
+
+templates.env.globals["csp_nonce"] = _csp_nonce
+
+
 def _login_redirect() -> RedirectResponse:
     return RedirectResponse("/login", status_code=303)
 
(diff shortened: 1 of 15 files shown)


GeiserX:
@coderabbitai review

PR #121, opened 2026-07-18
refactor: extract SSRF worker-URL validation into app/worker_proxy.py (CashPilot-sux)

GeiserX:
Splits a cohesive, self-contained slice out of the `main.py` god module — the architect-endorsed least-coupled candidate from bead `CashPilot-sux`.

## What moved

The SSRF/worker-URL guard → new **`app/worker_proxy.py`**:
- policy constants (`_WORKER_URL_POLICY`, metadata/loopback block sets, allowlist),
- `_parse_worker_allowlist`, `_normalize_ip`, `_assert_ip_not_blocked`, `_assert_ip_strict_allowed`,
- `_validate_worker_url` (DNS-rebinding-safe validation), `_pin_url_to_ip`.

`main.py`'s `_get_verified_worker_url` composes them and **stays** in main (it also needs `FLEET_API_KEY` + `database`). `worker_proxy` imports nothing from `main`, so **no new import cycle**. main.py drops ~160 lines and the SSRF policy is now isolated and unit-testable on its own.

## Test seams

- The 8 `app.main.socket.getaddrinfo` patches → `app.worker_proxy.socket` (that's where `_validate_worker_url` resolves the host).
- The strict-mode / allowlist tests that mutate the policy constants repoint their `main` alias to `app.worker_proxy` (where `_validate_worker_url` reads them).
- The `_validate_worker_url` / `_pin_url_to_ip` patches keep working via `main`'s import binding — unchanged.

## Verification

Behavior-preserving: **1172 tests green**; `ruff check` + `ruff format --check` clean; verified `worker_proxy` imports nothing from `app.main`.

## Scope

This lands the module-extraction half of `sux`. The other half — resolving the `main → routers → main` import cycle — is **intentionally left as-is**: it's benign and documented, and breaking it churns ~40 `app.main.*` test seams for no correctness gain (per the architect review). The bead can close on the extraction or stay open for the cosmetic cycle cleanup — your call.

Addresses bead CashPilot-sux (extraction half).

PR #177, opened 2026-08-02
refactor: break the main -> routers -> main import cycle (CashPilot-sux)

GeiserX:
The SSRF/worker-URL extraction this bead also called for [already shipped in #121]. What remained was the cycle — and the interesting part is how small the real coupling turned out to be.

The routers reached into `app.main` for `templates`, `client_ip`, the auth guards, `database`, `metrics`, `setup_token` and the login rate limiter. **Every one except the last was already a re-export** of something in `app.deps` or a plain module — `app.main` was just a pass-through.

So the routers now import their dependencies directly, and moving one module-level dict is what actually removes the cycle rather than relocating it.

`app.login_rate_limit` imports nothing from the app, so it can't be pulled back into a cycle later. Tests assert both: no router imports `main`, and the new module has no app-level imports.

## The seams were the real risk

`app.main` re-exports `_login_attempts` **by reference, not as a copy** — a conftest fixture clears `app.main._login_attempts` between tests. A copy would have left that fixture clearing an object nothing reads, and every later test would silently inherit the previous one's state. That's flakiness which surfaces much later and is very hard to place. A test pins the identity.

**One seam did legitimately move.** A test patched `app.main._check_login_rate`, which no longer intercepts anything now that the router calls the function where it lives — patching a name in `main`'s namespace never affected the other module. That test now patches `app.login_rate_limit.check_login_rate`: one explicit update, not a silent behaviour change.

## Two things the mechanical rename got wrong

Both caught by the suite rather than by reading:

- **`app.setup_token` collides with a `Form` parameter of the same name** inside `do_register`, so the bare import looked unused, was auto-removed, and the other use became a `NameError`. Now imported under an alias, with a comment saying why.
- The auth router's docstring still described the old arrangement.

## Behaviour unchanged

Tested as such: the window, the limit, per-address counting, and the deliberate plain dict (not a `defaultdict`) so an aged-out bucket is *removed* rather than leaking a permanent key for every address that ever hits `/login`.

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1749 passed**, coverage 94.66%

### DIFF
--- a/app/login_rate_limit.py
+++ b/app/login_rate_limit.py
@@ -1,0 +1,58 @@
+"""Login rate limiting (CashPilot-sux).
+
+Lifted out of ``app.main`` because it was the last thing the routers genuinely
+needed from it. Everything else they imported through ``app.main`` —
+``templates``, ``client_ip``, the auth guards — already lives in ``app.deps``
+and was only being re-exported, so moving this one module-level dict is what
+actually breaks the ``main -> routers -> main`` import cycle rather than merely
+moving it around.
+
+The state stays a module-level dict on purpose, and ``app.main`` re-exports the
+SAME object. Tests reach for ``app.main._login_attempts`` and a conftest fixture
+clears it between tests; rebinding it here — or handing out a copy — would leave
+those pointing at a dict nothing reads, and the fixture would silently stop
+isolating tests from each other.
+"""
+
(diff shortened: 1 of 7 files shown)


GeiserX:
@coderabbitai review

PR #178, opened 2026-08-02
refactor: componentize the duplicated app.js markup builders (CashPilot-cyc)

GeiserX:
> **Stacked on #176** (`guw`), which this depends on — the bead's other halves shipped in #120 (deploy-flow dedup) and #176 (inline handler removal).

Three copies of the env-input markup and two of the worker checkbox list, now **one builder each**. `app.js` loses 55 lines and, more usefully, loses what duplicated markup always does.

## It had already drifted

One env-input copy had quietly lost the label's `for` attribute **and** the description hint — so the same field rendered slightly less accessible in one of the three places it appears. Nobody would find that by reading any single copy. That's the actual cost of duplicated markup: it doesn't stay identical, it stays *almost* identical.

## The checkbox class stays a parameter — deliberately

This looks like leaving the flagged inconsistency in place, and it isn't. The setup wizard and the detail view can both be in the DOM at once, and their deploy handlers select checkboxes **by class**. A single shared class would make each one pick up the other's boxes and **deploy to workers the user never ticked**.

The duplication was the bug. The two class names are load-bearing.

## Verified, not inferred

Using the browser harness from #176:

- zero CSP violations, zero JS errors
- a delegated button still works end to end (wizard advances step-1 → step-2)
- full suite green

## Verification

- `pytest --cov=app --cov-fail-under=90` → **1767 passed**, coverage 94.65%


GeiserX:
@coderabbitai review

PR #180, opened 2026-08-02
fix: correct nine defects found auditing v1.10.0

GeiserX:
Nine defects from a post-release audit of v1.10.0. Each was verified against the live instance or a reproduction, not by reading.

## Earnings arithmetic

| Defect | Effect |
|---|---|
| `get_earned_by_platform` filtered `AND currency = 'USD'` | A fleet earning in MYST and GRASS showed a dashboard total near zero. Now converts where a rate exists; excludes only what cannot be priced. |
| `get_confirmed_payout_totals` used `COALESCE(fx_rate_usd, 1.0)` | Priced unrated payouts at parity — 500 MYST became $500. Unpriced rows now contribute nothing and the understatement is logged. Wrong in the safe direction. |
| `payouts.daily_rate` divided by reading count | Divided by the number of readings, not the days between them, so any gap in collection inflated the rate and pulled every payout projection early. Verified: 4.0 → 1.0 with a gap, unchanged for a clean series. |

## Access control

Three endpoints sat a rung below the repo's ladder (reads → auth, actions → writer, secrets → owner):

- `api_test_credentials` logs into a provider using the owner's stored credentials — was reachable by any authenticated viewer. Now **owner**.
- `api_confirm_payout` / `api_reject_payout` mutate and permanently delete payout history. Now **writer**.

## Recovery and operations

- **Recipient-mode backup verification was entirely broken.** `VerifyRequest` had no `recipient_private_key` field, so the handler always verified as the wrong identity and a correct backup reported as unverifiable. Verified `False` → `True`.
- A failed unpause was swallowed by `contextlib.suppress`, leaving a container paused and silently earning nothing. Now logs the container and the exact `docker unpause` command.
- The 401 lockout alarm fired on any auth failure rather than a run of them, so one transient blip raised it.
- `collector-live-check` treated pytest exit 5 (no tests selected) as failure, reddening the schedule for a non-event.

## Why the guard tests are here

A negative control: I removed `_require_auth_api` from five endpoints and the suite reported **2108 passed**. Every endpoint test patches the guard away before calling the handler, so nothing anywhere asserted that an anonymous caller is refused.

`tests/test_audit_guards.py` closes that, plus the shapes that hid the other defects: a Dockerfile.worker COPY closure check (a missing module builds fine and crash-loops only on a user's machine), route shadowing, and the split tariff key. Re-running the same control against it now yields exactly five failures and no collateral.

Two test files changed only to patch the guard the endpoint now enforces.

## Verification

`ruff check` + `ruff format --check` clean · **2130 passed** · coverage 94.79%

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected mixed-currency earnings and payout totals, including unrated currencies and USD parity.
  - Improved daily payout-rate calculations for sparse or irregular readings.
  - Enhanced backup recovery errors with manual recovery guidance.
  - Prevented uncertain container status from being reported as stopped or producing.

- **Security**
  - Tightened permissions for credential testing and payout actions.
  - Prevented encrypted worker credentials from appearing in responses.
  - Improved authentication and invalid backup-key handling.

- **New Features**
  - Added distinct payout alerts and restored notification support.

- **Compatibility**
  - Fleet economics supports current and legacy tariff settings.

### DIFF
--- a/.github/workflows/collector-live-check.yml
+++ b/.github/workflows/collector-live-check.yml
@@ -45,7 +45,11 @@ jobs:
           CASHPILOT_LIVE_CREDENTIALS: ${{ secrets.CASHPILOT_LIVE_CREDENTIALS }}
         run: |
           set -o pipefail
-          python -m pytest tests/ -m live -q 2>&1 | tee live.log
+          # Exit 5 is pytest's "no tests collected". No test carries the `live`
+          # marker yet, so without this the job fails EVERY night and files a
+          # false drift alert — which trains the maintainer to ignore the one
+          # alert designed to catch a real provider change.
+          python -m pytest tests/ -m live -q 2>&1 | tee live.log || [ "${PIPESTATUS[0]}" = "5" ]
 
       - name: Open an issue when a provider's shape changed
         if: failure()
--- a/app/database.py
+++ b/app/database.py
@@ -718,6 +718,80 @@ async def get_earnings_history(
         await db.close()
 
 
+def _usd_rate(currency: str, raw: Any) -> float | None:
+    """The USD price of one unit, or None when the reading cannot be priced.
+
+    USD is parity BY DEFINITION — trusting a stored rate on a USD row would let
+    a bad rate rewrite money the collector reported exactly. For anything else
+    only a finite positive number is a price:
+
+    * ``0`` reports the platform as having earned nothing, which is
+      indistinguishable from a genuinely flat balance;
+    * a negative defeats the payout clamp, which applies to the delta while the
+      sign is applied after it, so the total comes out NEGATIVE;
+    * ``inf``/``nan`` poison the total outright.
+
+    Anything rejected here is treated as unpriced — excluded and reported —
+    rather than believed.
+    """
+    if currency == "USD":
+        return 1.0
+    if raw is None:
+        return None
+    value = float(raw)
+    return value if math.isfinite(value) and value > 0 else None
+
+
+async def _usd_earned_per_date(db: Any) -> tuple[dict[str, float], int]:
+    """USD earned per calendar date, summed across platforms.
+
+    ONE implementation behind both the dashboard cards and the trend chart.
+    They previously carried separate copies of the arithmetic and both filtered
+    ``currency = 'USD'``, so an installation earning MYST, GRASS or ANYONE saw
+    "$0.00 today", "$0.00 this month" and a flat-zero chart while its balances
+    climbed — verified against a MystNodes-only fixture, which reported 0.00
+    where the correct figure was 0.20.
+
+    Same rules as :func:`get_earned_by_platform`, and deliberately so:
+
+    * the delta is taken in the NATIVE currency and only then priced, because a
+      balance is a running total and converting before subtracting puts the
+      movement of the exchange rate inside the earnings figure;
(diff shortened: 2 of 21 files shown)


GeiserX:
Both findings were correct and are fixed in 5c4498f. Notes on each:

**`app/database.py` — delta before conversion.** Confirmed, and it was a regression introduced by the previous commit in this same PR rather than an existing bug: before it, non-USD rows were dropped by `AND currency = 'USD'`, so no rate could contaminate anything. Making them count re-introduced the ordering problem.

The delta is now taken in the native currency and priced at the later reading's rate. Two consequences of the reordering, both suggested and both implemented:
- a USD row is parity regardless of any `fx_rate_usd` stored against it;
- an unpriced reading now *drops* the baseline rather than anchoring the next delta, so an unpriced period is not silently counted whole by a later differencing.

Rate-movement regression tests added in `tests/test_power.py` against a real SQLite database. Negative control: restoring the convert-then-delta arithmetic fails exactly the three tests that target it (falling rate, rising rate, currency switch) and no others.

**`app/orchestrator.py` — unpause failure.** Confirmed, including the consequence you traced through `worker_api`. It now re-raises when the reads succeeded, and still suppresses when a read error is already propagating, so the real cause is never replaced by a downstream symptom. Losing an archive is recoverable by taking it again; an unnoticed paused earner just stops paying.

**A third defect, found while writing the tests you asked for.** The elapsed-days divisor added earlier in this PR returned a negative number for out-of-order dates, which would invert the payout rate and project into the past. Fixed with `abs()`; `app/payouts.py` is now at 100% line coverage.

Every line this PR adds is covered — verified by cross-referencing the added-line set against the coverage report, not by the project total.

GeiserX:
@coderabbitai review

GeiserX:
Correct, and worse than the two cases the previous guard did catch — verified, `float("inf")` returned an **infinite** platform total.

Widened the production check to `math.isfinite(rate) and rate > 0` rather than special-casing infinity, and added both the `inf` and `nan` regression cases you suggested.

On `nan` specifically: it is currently unreachable *through the database*, because SQLite stores NaN as NULL and the unpriced path already handles NULL. I kept the test anyway and noted why in it — the guard should not silently depend on that storage detail staying true.

The `0.0001` case is still there to keep the guard from over-reaching onto legitimate small token rates.

GeiserX:
Right, and it was a defect in my own fix from the commit before — thank you.

`record_probable_payout` returns `None` while one is already pending, so `_detect_payout` reported the payout on the run that noticed it and `None` on every run after. Rebuilding the bell from that alone made the question vanish on the next collection, before the user could answer it.

Fixed by rebuilding from the pending rows in the database rather than from what was noticed in the last few seconds — those rows are the actual source of truth for "there is still a question outstanding" — deduped against anything already detected in the current run.

I also fixed the mirror-image problem your finding implies: **answering** a payout now retires its stored alert. Without that, the alert row survives and `_warm_collector_alerts` restores it on the next restart, asking again about a payout the user already settled. The reject path reads the platform *before* the row is deleted, since afterwards there is nothing to look it up from.

Six tests added, including the no-duplicate case and a database-failure case (a broken lookup must not take down the collection run).

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #181, opened 2026-08-03
feat: make payouts visible — a queue to answer them, and progress toward the next one

GeiserX:
First slice of the 1.11.0 UI work. v1.10.x shipped **35 `/api/**` endpoints with no frontend consumer** — features designed, implemented, tested, and then invisible. This wires up the one with a real cost attached.

## Why this one first

A balance drop is recorded as a **probable** payout. Until someone answers "yes, I was paid", it never counts toward lifetime earnings. There was nowhere to answer.

So a real payout looked exactly like a loss — permanently — which is the opposite of what the feature was built for. `GET /api/earnings/payouts` and the confirm/reject endpoints existed, were tested, and were called by nothing.

## What it does

A dashboard card, **hidden until there is something to answer** — an empty card every day trains people to ignore the one day it matters.

- Confirm is one click.
- Reject asks first: it is a hard `DELETE` with no undo.
- A failed fetch leaves the card alone rather than hiding it. Unknown is not "nothing pending", and hiding it would silently drop a question the user still owes an answer to.

## Verified in a real browser, not by reading

Driven over CDP against a seeded instance:

```
BEFORE click: cardVisible=true  rows=1  platform=mysterium
              detail="Balance dropped by 120.50 MYST on 2026-08-03"
              confirmBtns=1  rejectBtns=1  inlineOnclick=0  handlersExist=true
AFTER click : cardVisible=false  rows=0
csp violations: none
js errors     : none
RESULT: PASS
```

**The browser check earned its keep immediately.** The first run rendered a **24.90 USD** payout as **"£18.55"** — the dashboard's display currency applied to it. Everywhere else that is correct; here it is a converted approximation of one specific real transaction that the user is about to check against the provider's own page. The native amount now leads, the conversion follows marked `≈`, and it is suppressed when the two are identical. No string test would have caught that.

## Tests

`tests/test_frontend_wiring.py` guards all four pieces of the wiring, each of which fails silently on its own: the template anchor, the CP export (an unexported handler is a dead button — `delegate.js` resolves `data-action` against `CP`), the call from `loadDashboard`, and the absence of inline handlers.

They also carry their own correction: the first version sliced a fixed number of characters out of `app.js` and read on into the next function, so it reported that `confirmPayout` asks for confirmation when what it had actually found was `rejectPayout`. Now bounded on the next function.

Negative control: removing the three CP exports fails `TestEveryDeclaredActionExists`.

## Verification

`ruff check` + `ruff format --check` clean · **2214 passed** · coverage 95.02%

Not merging without your OK.

## Summary by CodeRabbit

* **New Features**
  * Added payout queues with confirmation/rejection actions, progress tracking, and clearer balance details.
  * Added credential health, service-specific guidance, fleet running costs, per-node earnings, and improved currency handling.
  * Added deployment preflight warnings and service deploy-risk notices.
  * Added setup-token guidance and validation during account registration.

* **Bug Fixes**
  * Improved payout, service-detail, and fleet layouts, status readability, and cost-attribution explanations.
  * Preserved chart data when earnings loading fails and retained USD labels when conversion rates are unavailable.
  * Improved external-link safety, onboarding errors, and wizard service selection.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'Dockerfile'
+      - 'entrypoint.sh'
       - 'Dockerfile.worker'
       - 'pyproject.toml'
       - 'uv.lock'
@@ -72,7 +73,7 @@ jobs:
           BUILD_UI=false
           BUILD_WORKER=false
 
-          if echo "$CHANGED" | grep -qE '^(Dockerfile|pyproject\.toml|uv\.lock|app/(main|database|catalog|auth|compose_generator|exchange_rates|constants|collectors|templates|static|routers|deps|metrics|setup_token))'; then
+          if echo "$CHANGED" | grep -qE '^(Dockerfile|entrypoint\.sh|pyproject\.toml|uv\.lock|app/(main|database|catalog|auth|compose_generator|exchange_rates|constants|collectors|templates|static|routers|deps|metrics|setup_token))'; then
             BUILD_UI=true
           fi
           if echo "$CHANGED" | grep -qE '^services/'; then
@@ -80,7 +81,7 @@ jobs:
             BUILD_WORKER=true
           fi
 
-          if echo "$CHANGED" | grep -qE '^(Dockerfile\.worker|pyproject\.toml|uv\.lock|app/(worker_api|orchestrator|constants|catalog|fleet_key)\.py)'; then
+          if echo "$CHANGED" | grep -qE '^(Dockerfile\.worker|entrypoint\.sh|pyproject\.toml|uv\.lock|app/(worker_api|orchestrator|constants|catalog|fleet_key)\.py)'; then
             BUILD_WORKER=true
           fi
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,27 @@ jobs:
         with:
           python-version: "3.14"
 
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+
+      # Nothing here parsed the JavaScript, and pytest cannot: appending a
+      # syntax error to app.js leaves the whole suite green. What ships in that
+      # state is a completely dead dashboard — app.js is one IIFE assigned to
+      # CP, so a parse failure anywhere means CP is never defined and every
+      # data-action button silently does nothing. No error page, no failed
+      # request, just a UI where nothing works.
+      - name: Parse the JavaScript
+        run: |
+          for f in app/static/js/*.js; do node --check "$f"; done
+
+      # Behaviour pytest cannot reach. This one needs no browser: it runs the
+      # real currency functions against controlled exchange-rate state, and
+      # exists because formatCurrency once labelled an UNCONVERTED USD amount
+      # with the viewer's currency symbol — $24.90 rendered as "£24.90".
+      - name: Currency formatting behaviour
+        run: node scripts/currency_check.mjs
+
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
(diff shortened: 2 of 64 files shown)


GeiserX:
Both correct, and the first one led somewhere worse than the symptom.

**Duplicate amount.** Right about the cause — comparing `$24.90` against `24.90 USD` finds them different even when nothing was converted. But comparing against `_displayCurrency` instead is still wrong: when that currency has **no rate**, the amount stays in USD while the preference says GBP. The decision now asks which currency the formatter will *actually* use, which also covers an unpriced token staying in its own units.

**The larger defect underneath.** Chasing it in a browser against a freshly restarted server, `formatCurrency` rendered **$24.90 as "£24.90"** — it left the amount unconverted (no GBP rate yet) and then formatted it *as* GBP. Not a conversion; the same number wearing a different sign. Rates load asynchronously, so this hit **every figure on the dashboard on every page load** until they arrived, and permanently for any currency with no rate. Fixed: the label now follows the conversion that actually happened.

**My test was worse than the bug.** `test_it_does_not_repeat_itself_when_both_are_the_same` asserted the literal source text `converted !== native` — the implementation, not the behaviour — so it passed while that exact comparison produced the duplication it claimed to prevent. Replaced, and the real proof is now `scripts/currency_check.mjs`, which runs the actual functions against controlled rate state across five cases:

```
USD payout, USD dashboard           approx=false  ($24.90)
USD payout, GBP dashboard + rate    approx=true   (£18.55)
USD payout, GBP dashboard, NO rate  approx=false  ($24.90)   <- was "£24.90"
MYST payout, no crypto rate         approx=false  (24.90 MYST)
MYST payout, crypto rate, USD dash  approx=true   ($2.17)
```

**Role gating.** Also correct — `_canWrite` now gates both buttons, following the idiom the deploy button already uses, with a line explaining why rather than an unexplained empty row.

2231 passed, coverage 95.02%.

GeiserX:
Both fixed — the second one somewhere other than where you pointed, for a reason worth recording.

**`api_per_node_earnings` unhandled exception.** Correct and fixed. That call reaches a third-party API that times out, rate-limits and returns HTML on a bad day; none of that is a CashPilot fault and none of it should arrive as a raw 500. It now logs and returns an empty breakdown, matching how every other collector call in the file degrades. The collector is still closed on that path, so a flaky provider cannot leak a session per request.

**`rel="noopener noreferrer"` — right risk, wrong location, and undercounted.**

It is **13 files, not 5** — every migrated credential hint carries a `target='_blank'` anchor, not just the ones in the diff you sampled.

More importantly, **adding `rel` to the YAML would have done nothing.** `sanitizeHint` (app/static/js/app.js) strips every attribute it does not explicitly keep — only `href` and `target` survive on an anchor. A `rel` written into a service file is removed on its way to the DOM. The fix would have looked applied, passed review, and had no effect.

So it is set by the sanitiser, *after* the stripping loop. That covers all 13 hints and every future one, and cannot be forgotten by whoever writes the next.

Verified in a real DOM against the real shipped hint (`scripts/hint_sanitizer_check.mjs`, committed):

```
shipped   : target=_blank  rel=noopener noreferrer  href=https://earnapp.com
yamlRel   : noopener noreferrer   <- a YAML-written rel is OVERWRITTEN, not honoured
noTarget  : null                  <- a link without target gains nothing
dangerous : href=null  onclick=null  scripts=0   <- still stripped
```

On severity: modern browsers imply `noopener` for `target=_blank`, so this was defence in depth rather than a live hole. `noreferrer` is the part still worth having — these links point at a provider dashboard, and the referrer would otherwise name the user's CashPilot host.

The two currency/`_canWrite` comments were already fixed earlier in this PR (commit 5c4498f onward); they were flagged against older commits in the stack.

2291 passing, coverage 95.07%.

PR #182, opened 2026-08-03
fix: ten verified audit findings (beads batch 1)

GeiserX:
First of the bead batches. **Ten beads, all independently re-verified against this branch before being touched** — the audit's adversarial pass had already refuted 13 of 94 findings, so nothing here was taken on trust.

Targets `feat/payout-queue-ui`, not `main`, deliberately: `release.yml` fires on every push to main touching `app/**`, so merging batches straight to main would cut one release each. Stacking here yields a single release at the end.

| Bead | Defect |
|---|---|
| `c1q` + `i21` | Upgraded volumes reported every credential "fresh", including ones expired days earlier |
| `153` | `bool("false")` is True — the dedicated flag gave opposite advice |
| `a02` | Legacy tariff key worked on one endpoint, ignored on the other |
| `4ks` | `/docs`, `/redoc`, `/openapi.json` served unauthenticated |
| `0id` | One of two credentials showed "Configured" while collection silently never ran |
| `0rw` | Credential hints never reached the Settings page they were written for |
| `x2r` | Changing `entrypoint.sh` published no image |
| `w3g` | Startup refused with a supplied encryption key, telling the user to supply one |
| `ecc` | Salad hint pointed at a host that does not resolve |
| `i5l` | `CLAUDE.md` documented a synthetic baseline row that does not exist |

`c1q`/`i21` are the same defect filed from two audit areas — one fix closes both.

**Verification:** 2386 tests, 95.08% coverage, ruff clean, all JS parses. Negative controls run: reverting the docs and dedicated fixes turns four tests red.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved collector setup status to clearly identify incomplete credentials.
  - Sanitized credential guidance is now displayed safely in settings.
  - Preserved supplied encryption keys across restarts when persistence is unavailable.
  - Corrected tariff handling, boolean settings, and first-reading earnings behavior.
  - Updated Salad connection instructions to use the correct website.
  - Disabled interactive API documentation and schema endpoints for improved security.

- **Documentation**
  - Clarified earnings baseline behavior and updated collector setup guidance.

- **Release Improvements**
  - Changes to the application entrypoint now trigger the required builds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'Dockerfile'
+      - 'entrypoint.sh'
       - 'Dockerfile.worker'
       - 'pyproject.toml'
       - 'uv.lock'
@@ -72,7 +73,7 @@ jobs:
           BUILD_UI=false
           BUILD_WORKER=false
 
-          if echo "$CHANGED" | grep -qE '^(Dockerfile|pyproject\.toml|uv\.lock|app/(main|database|catalog|auth|compose_generator|exchange_rates|constants|collectors|templates|static|routers|deps|metrics|setup_token))'; then
+          if echo "$CHANGED" | grep -qE '^(Dockerfile|entrypoint\.sh|pyproject\.toml|uv\.lock|app/(main|database|catalog|auth|compose_generator|exchange_rates|constants|collectors|templates|static|routers|deps|metrics|setup_token))'; then
             BUILD_UI=true
           fi
           if echo "$CHANGED" | grep -qE '^services/'; then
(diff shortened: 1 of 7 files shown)


GeiserX:
@coderabbitai review

GeiserX:
Both addressed — both were right.

**Persistence-failure message.** The fix stopped treating an env-supplied key as ephemeral but left the error text claiming credentials "will be unreadable after a restart". That is untrue while `CASHPILOT_ENCRYPTION_KEY` is set, and it is the kind of false alarm that trains people to ignore the real one. A **minted** key keeps the loud error; a **supplied** key now gets a warning stating what actually matters — keep the variable set, because without the cached file it is the only copy.

**Migration comment.** It still described the back-fill this PR removes. Left alone it would have invited someone to restore exactly the behaviour being deleted. Now points at the note below it.

Also worth flagging: this PR and #183–#186 were **skipped** by auto-review — "auto reviews are disabled on base/target branches other than the default branch" — because they stack on `feat/payout-queue-ui` so the batches produce a single release rather than one per PR. The check showed green with no review behind it. I've requested reviews on all of them manually.

2386 tests, ruff clean.

PR #183, opened 2026-08-03
fix: unknown balances stop rendering as zero (beads batch 2)

GeiserX:
Eight beads, **two fixes** — three audit areas filed the same two defects between them, so the duplicates collapse.

| Beads | One defect |
|---|---|
| `vp6` + `ikh` + `7qk` | `balance_map.get(slug, 0.0)` — a service never read showed `$0.00` |
| `3oa` + `jkd` + `s2b` | Payout card asserted 0.00 lifetime and a 0% bar for a service never read |

**Why this one hid so well.** For a never-read service the dashboard showed *Running*, a balance of *$0.00*, and a bell saying *All collectors healthy* — three independent signals agreeing on something none of them had checked. `collector_disconnected` needs an alert that was never raised, `collector_needs_setup` only inspects config keys, and the flatline detector explicitly skips rows whose max balance is 0.

**A trap worth noting:** the renderer's `const balance = ... || 0` would have coerced the new `null` straight back to a confident zero — undoing the fix while every test still passed. Removed.

**One self-correction.** I also nulled `projection`; an existing test objected and it was right. `payouts.project` already returns `NOT_ENOUGH_DATA`, which says *why* there is no estimate — strictly better than null. The misleading part was the card drawing a 0% bar from it, fixed where the card is drawn.

**Verification:** 2357 tests, 95.07% coverage, ruff clean, JS parses. Negative control: restoring the `0.0` default turns the API test red.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1037,7 +1037,21 @@ async def api_services_deployed(request: Request) -> list[dict[str, Any]]:
             "slug": slug,
             "name": svc["name"] if svc else slug,
             "container_status": agg["best_status"],
-            "balance": balance_map.get(slug, 0.0),
+            # None, not 0.0, when CashPilot has never read this service.
+            #
+            # A missing earnings row means "no reading", and rendering it as
+            # $0.00 told the user a service was running and earning nothing —
+            # when the truth was that nothing had ever looked. Neither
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #184, opened 2026-08-03
fix: API shape changes stop being recorded as zero earnings (beads batch 3)

GeiserX:
Seven beads. Two themes: an upgrade that could brick startup, and collectors that recorded a provider's API change as real earnings of zero.

**`y5t` — an interrupted upgrade bricked startup permanently.**

`executescript` commits per statement, and the workers rebuild had no `DROP IF EXISTS` and no transaction. Interrupt it between CREATE and RENAME and `workers_new` survives; the guard above is still true next boot, so the rebuild re-runs and dies on "table already exists". Reproduced on a seeded pre-`client_id` volume:

```
OperationalError: table workers_new already exists
```

That repeats on every restart — the app never comes back. Now re-entrant, worker rows preserved.

**`3yn`/`rtb` (Storj), `10d` (Anyone), `fzw` (Grass) — a renamed field became $0.00.**

All three used `.get(key, default)`, so a shape change produced a confident zero written as a real balance: no error, so no alert; and the flatline detector explicitly skips rows whose max balance is 0.

Storj is worst — its zero is a *drop* from the previous reading, so the payout detector asked the user to confirm receiving their entire balance. Money never paid, and **a confirmed payout is permanent**.

Each now separates an absent key from a genuine zero. `"0"`/`"null"` still mean zero, an empty device list still means zero, Storj still sums whichever fields are present.

**`0ja`** — `/api/preferences` spawned a full fleet collection for any authenticated user, while `/api/collect` gates the same call behind `_require_writer`. The preference still saves for a viewer; only the collection is gated.

**`dbh`** — both Deploy buttons were enabled for writers while `api_deploy` requires owner, so a writer typed in provider credentials and *then* got a 403.

**Verification:** 2367 tests, 95.13% coverage, ruff clean, JS parses. Negative controls: reverting the migration guard fails 3 tests, reverting the collection gate fails 1.

### DIFF
--- a/app/collectors/anyone.py
+++ b/app/collectors/anyone.py
@@ -66,7 +66,18 @@ class AnyoneCollector(BaseCollector):
             logger.debug("No messages returned for fingerprint %s", fingerprint)
             return 0.0
 
-        data = messages[0].get("Data", "0")
+        # A missing Data key is a shape change, not zero rewards. Defaulting
+        # to "0" made an unparseable AO contract response indistinguishable
+        # from a relay that genuinely earned nothing — and the resulting 0.0
+        # was written as a real reading with no error, so no alert, no
+        # flatline warning (that detector skips zero balances), and no way for
+        # the user to tell.
+        if "Data" not in messages[0]:
+            raise ValueError(
+                f"AO response for {fingerprint} has no Data field "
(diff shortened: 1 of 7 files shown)


GeiserX:
@coderabbitai review

PR #185, opened 2026-08-03
fix: totals that were wrong, not merely missing (beads batch 4)

GeiserX:
Four beads. Three totals that misled in different ways.

**`47t` — the dashboard Total silently dropped whole holdings.**

`to_usd` consults only the *live* caches, so a crypto whose rate lookup was merely stale vanished from the headline figure. Measured on a seeded run:

```
before   110 MYST, no live rate  ->  total 0.00   (holding dropped entirely)
after    110 MYST, no live rate  ->  total 13.20  (stored rate 0.12)
```

Not a new rule: `get_earnings_dashboard_summary` and `get_earned_by_platform` already price rows this way, and `upsert_earnings` stores `fx_rate_usd` precisely so historical values "cannot be reconstructed later" without it. The Total was applying a *stricter* rule than the cards beside it — which is why they disagreed. It now uses the same `_usd_rate` validator, so 0/negative/inf/nan are still rejected.

The query didn't even select `fx_rate_usd`, so the value was unreachable from this path.

**`oj4`** — what still can't be priced is now reported as `unpriced_platforms` instead of logged and forgotten. A total that silently omits holdings looks exactly like a correct one.

**`7oj`** — `month_change` was a hardcoded `0.0` nothing computed, rendered as **"+0.0%"** in the positive style, permanently. Now null, and the renderer shows blank rather than throwing.

**`he1`** — fleet net subtracted the cost of *some* machines from the gross of *all* of them, so the more machines you couldn't price, the healthier the fleet looked. Its own docstring promises it "never quietly understates what the fleet costs."

**Verification:** 2363 tests, 95.08% coverage, ruff clean, JS parses. Negative controls: reverting `he1` and `7oj` turns three tests red.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -769,7 +769,12 @@ async def get_earnings_summary() -> list[dict[str, Any]]:
     try:
         cursor = await db.execute(
             """
-            SELECT platform, balance, currency, date
+            -- fx_rate_usd is selected because the dashboard total needs a
+            -- fallback when no LIVE rate is cached. Without it a crypto
+            -- balance whose rate lookup is merely stale gets dropped from the
+            -- headline figure entirely, even though the rate it was recorded
+            -- at is sitting right here in the row.
+            SELECT platform, balance, currency, date, fx_rate_usd
             FROM earnings
             WHERE (platform, date) IN (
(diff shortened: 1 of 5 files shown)


GeiserX:
@coderabbitai review

PR #186, opened 2026-08-03
fix: session hijack when adding a user, and unconfirmed worker keys (beads batch 5)

GeiserX:
Four beads.

**`5e6` — an owner adding a user was logged into that user's account.**

`do_register` minted a session for the account it had just created and set it as the *caller's* cookie. An owner adding a viewer was silently demoted into that viewer's session — their own access gone, with no sign anything had happened beyond landing on a page with fewer controls. Only the first-run owner is signed in now.

**`zcc` — the UI signed worker commands with a key the worker never confirmed.**

`get_worker_key` discards the `confirmed` flag that `get_worker_key_state` returns. The unconfirmed state is reachable whenever a worker fails to persist its key and logs *"staying on shared key"*. The worker then reads as **online** from its heartbeats while every deploy, start, stop, restart and logs call fails **401** — two symptoms with nothing connecting them on screen.

**`cxy`** — per-worker watts/dedicated settings were keyed on the AUTOINCREMENT row id. `delete_worker` removes the row and nothing else, so a host removed and re-enrolled came back under a new id, orphaning its settings while the old keys lingered. Now keyed on `client_id`, with the old key still honoured.

**`153`** (this branch's half) — same helper fixes `bool("false") == True` on TEXT config values.

**On review coverage:** CodeRabbit had skipped *all* these PRs — "auto reviews are disabled on base branches other than the default" — a side effect of stacking them on the integration branch. The check showed green with zero review. I've triggered reviews on #182–#185 manually.

**Verification:** 2351 tests, 95.08% coverage. Negative controls for both behaviour changes — and the first `5e6` control **passed**, because my revert patch silently didn't match. Redone against the real text, it fails the two tests that name the defect.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1555,8 +1555,17 @@ async def _get_verified_worker_url(worker: dict[str, Any]) -> tuple[str, dict[st
     # shared bootstrap key only for workers that have not enrolled yet. Post-cutover
     # an enrolled worker rejects the shared key, so the UI must present its own.
     cid = worker.get("client_id") or ""
-    worker_key = await database.get_worker_key(cid) if cid else None
-    auth_key = worker_key or FLEET_API_KEY
+    # Use the per-worker key only once the worker has CONFIRMED it.
+    #
+    # get_worker_key discards the confirmed flag, so the UI signed outbound
+    # commands with a key the worker may never have received — reachable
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #187, opened 2026-08-04
fix: credentials alone now collect, and the collection test can fail (beads batch 6)

GeiserX:
Targets `main` directly this time — the integration branch is merged and released, and CodeRabbit skips PRs that don't target the default branch.

**`1f8` + `ahh` — the biggest user-facing gap in the audit.**

Settings promises: *"You don't need to deploy containers through CashPilot — just add your credentials."* That was false for **12 of the 15 collectors**.

Collection is driven by deployment rows — `make_collectors` iterates deployments — and `api_set_config` created that row only for services with **no Docker image** (three of them). For everything else the badge flipped to "Configured" and no reading ever arrived. That's the entirety of user story B, silently doing nothing.

Verified end to end, not argued:

```
saving honeygain credentials  ->  rows: [('honeygain', 'external')]
                                  collectors: ['honeygain']
before the fix                ->  rows: []   collectors: []
```

The teardown moved with it and now keys off the deployment's **status**, not the image. Clearing credentials removes the placeholder and never touches a row with a real container id — gating on the image would have orphaned a running container from the dashboard.

**`55m` + `rhn` — `test_run_collection` had no assertions at all.**

It ran the function and checked nothing, so the whole success path could be deleted and the suite stayed green. It now asserts the reading is stored with the collector's values, the payout branch is reached, and what `fx_rate_usd` is: live rate when priced, `1.0` for USD by definition, `None` — never parity — when unpriced.

**Mutation-tested.** Deleting the upsert fails 3 tests; forcing `fx_rate_usd` to parity fails 2. My first attempt at that second mutation *passed* — there are two call sites and I'd patched the one in `_detect_payout` instead of the collection path.

**Verification:** 2452 tests, ruff clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Saving complete collector credentials now tracks services without deployed containers, including image-based services.
  - Clearing credentials removes only tracking placeholders and preserves active container deployments.
  - Improved handling of incomplete credentials and services without container images.
  - Collection processing now more reliably preserves earnings, detects payouts, handles currency conversion, and reports unpriced currencies correctly.

- **Tests**
  - Expanded coverage for credential, deployment-tracking, collection, payout, and currency scenarios.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2956,14 +2956,22 @@ async def api_set_config(request: Request, body: ConfigUpdate) -> dict[str, str]
         svc = catalog.get_service(slug)
         if not svc:
             continue
-        docker_conf = svc.get("docker", {})
-        has_image = bool(docker_conf and docker_conf.get("image"))
-        if has_image:
-            continue  # Docker services get deployed normally
+        # No has_image gate. Collection is driven by DEPLOYMENT ROWS
+        # (make_collectors iterates deployments), so a service with no row is
+        # never collected no matter how complete its credentials are. Skipping
+        # image-backed services here meant that for 12 of the 15 collectors,
+        # saving credentials did nothing at all — while Settings promised
+        # "You don't need to deploy containers through CashPilot — just add
+        # your credentials." The badge flipped to Configured and no earnings
+        # ever arrived.
+        #
+        # Creating the row is safe for a user who later deploys through
+        # CashPilot: api_deploy calls save_deployment with the real container
+        # id and status, replacing this placeholder.
         existing = await database.get_deployment(slug)
         if not existing:
             await database.save_deployment(slug=slug, container_id="", status="external")
-            logger.info("Auto-created external deployment for %s", slug)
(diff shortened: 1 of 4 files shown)


GeiserX:
Good catch — that was a real hole, not a theoretical one.

`set_config_bulk` upserts, so a credential set can legitimately arrive across several requests. Judging completeness against the request payload alone meant neither request ever saw a complete set. Reproduced before fixing:

```
save {honeygain_email}     -> config has email
save {honeygain_password}  -> config has BOTH keys
deployments                -> []          <- no row, collection still dead
```

Completeness now reads the merged config. I kept the scan scoped to the slugs the request touched — widening it to the whole catalog was the obvious wrong fix, since it would silently enable collection for a service configured long ago and since abandoned. That's asserted too.

```
after 1st (partial)  : []
after 2nd (complete) : [('honeygain','external')]
unrelated partial    : [('honeygain','external')]   <- undisturbed
```

Negative control: reverting to request-only completeness fails the split-request test. 2456 tests, ruff clean.

PR #189, opened 2026-08-04
fix: releases that publish nothing, and docs naming the wrong encryption key (beads batch 7)

GeiserX:
**`gn6` + `l7t` — a release could go green having published nothing.**

`release.yml` decided what to build from two hand-maintained regexes, both drifted from the Dockerfiles they mirror.

The UI regex named **13 of 27 modules**, while the UI image does `COPY app/ ./app/` — everything. A change to any of the other 14 (`payouts`, `preflight`, `power`, `egress`, `lan_isolation`, `notify`, …) set `BUILD_UI=false` → version step skipped → `new_tag` empty → tag, Release and build job all skipped. **Skipped steps don't fail a run**, so it reported success.

The worker regex omitted `egress.py` and `state_backup.py` — both COPY'd into the worker image and imported by `worker_api` at runtime. With `build_worker=false` the pipeline **retags the previous image**, so the worker ships under a new version tag containing old code, and `verify-tags` only runs `docker manifest inspect`, which a retag satisfies.

Neither is restated now: any `app/` change builds the UI, and the worker list is parsed out of `Dockerfile.worker` at run time. A test asserts every module `worker_api` imports is actually in the worker image — the contract that decides whether a miss is a crash on the user's machine or a build failure here.

**`dxi` — six places named the wrong key, harmfully.**

They said `CASHPILOT_SECRET_KEY` encrypts credentials. It signs sessions; `CASHPILOT_ENCRYPTION_KEY` is the Fernet key at `/data/.fernet_key`. Follow the old advice and you believe your credentials are portable, never back up the key file, and lose every one the first time the volume is recreated. README already said this correctly — the other six contradicted it.

Fixed all six, named the key file so it can be backed up, and added the variable to the Unraid template (masked), since that's the whole configuration surface for those users.

**Verification:** 2476 tests, 95.13% coverage. Negative controls: narrowing the UI trigger fails 2 tests; restoring the `index.md` claim fails 2 more.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that `CASHPILOT_SECRET_KEY` signs login sessions.
  * Added guidance for `CASHPILOT_ENCRYPTION_KEY`, which encrypts stored credentials.
  * Documented credential key storage, backup, recovery, and configuration options.
  * Updated the Unraid template with separate session-signing and encryption settings.

* **Bug Fixes**
  * Improved release detection so UI and worker builds trigger for all relevant application and module changes.

* **Tests**
  * Added coverage verifying release triggers, Dockerfile module packaging, and encryption-key documentation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,14 @@ jobs:
           BUILD_UI=false
           BUILD_WORKER=false
 
-          if echo "$CHANGED" | grep -qE '^(Dockerfile|entrypoint\.sh|pyproject\.toml|uv\.lock|app/(main|database|catalog|auth|compose_generator|exchange_rates|constants|collectors|templates|static|routers|deps|metrics|setup_token))'; then
+          # The UI image does `COPY app/ ./app/` — the WHOLE directory — so ANY
+          # change under app/ belongs in a UI build. The hand-maintained module
+          # allowlist that used to live here named 13 of 27 modules, so a change
+          # to any of the other 14 (payouts, preflight, power, egress,
+          # lan_isolation, notify, ...) produced a GREEN release run that
+          # published nothing at all: no tag, no GitHub Release, no images.
+          # Skipped steps do not fail a run, so it looked like a success.
+          if echo "$CHANGED" | grep -qE '^(Dockerfile|entrypoint\.sh|pyproject\.toml|uv\.lock|app/)'; then
             BUILD_UI=true
           fi
           if echo "$CHANGED" | grep -qE '^services/'; then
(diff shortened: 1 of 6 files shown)


GeiserX:
Fixed, and you were right about where it bites.

`app/database.py` keeps the stored key and logs a warning when `CASHPILOT_ENCRYPTION_KEY` differs — switching keys would make every existing credential unreadable. So "overridable" is wrong precisely in the restore case: someone putting their backup key on an instance that still has a stale `/data/.fernet_key` would expect it to take effect, and it is ignored.

All three docs now state it is adopted **only when that file is absent** — which is also the reason setting it on a healthy instance is safe rather than dangerous.

Pinned with a test that rejects "overridable" on any line naming the variable and requires all three docs to state the precedence, so the imprecision cannot come back quietly.

2478 tests, ruff clean.

PR #190, opened 2026-08-04
fix: an unreachable machine is not a machine earning nothing (beads batch 8)

GeiserX:
Two findings, one mistake: both endpoints are built only from **online** workers, and both then presented the resulting absence as measured fact.

**`daq` — an offline worker was told to switch itself off.**

`per_worker_gross` comes from online workers only, and the endpoint defaulted a missing entry to `0.0`. So a host that had merely stopped heartbeating produced a confident financial recommendation:

> geiserback earns about 0.00 a month and costs about 9.49 in electricity — roughly 9.49 out of pocket. Since this machine runs only these services, turning it off would save that.

Its earnings were also silently reattributed to whatever workers were still reporting, so the rest of the fleet looked better than it was at the same moment.

`assess_machine` now takes `monthly_gross | None` and returns `UNKNOWN` with no cost, net or gross. A machine that genuinely earned `0.00` is **still judged** — asserted, because a fix that stopped judging anything would pass the other tests.

**`1qy` — an unreachable worker read as "you have no services".**

Three minutes after a host stops heartbeating, the table emptied and the dashboard stated as fact that the user had nothing and should start over — with a Setup Wizard button. The containers were still running and still earning.

The empty state now asks `/api/workers` before choosing its sentence, says containers keep running and earning while a worker is offline, and when it cannot reach that endpoint either, asserts **neither** case rather than guessing. A genuinely empty install still gets the wizard.

**Verification:** 2493 tests, 95.13% coverage, ruff clean, JS parses. Negative controls: removing the unknown-gross branch fails 3 tests; collapsing the empty-state condition fails 1.

### DIFF
--- a/app/machine_economics.py
+++ b/app/machine_economics.py
@@ -81,7 +81,7 @@ def break_even_price(monthly_gross: float, watts: float) -> float | None:
 def assess_machine(
     *,
     name: str,
-    monthly_gross: float,
+    monthly_gross: float | None,
     watts: float | None,
     price_per_kwh: float | None,
     metered: bool = True,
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #191, opened 2026-08-04
test: guards that could not fail (beads batch 9)

GeiserX:
Two tests that existed, passed, and protected nothing.

**`1xg` — `test_offline_workers_skipped` fed a worker with no containers.**

Deleting the `status != "online"` skip still produced `[]`, so the test passed while asserting nothing about the rule it is named for. The offline worker now has a container, so removing the skip leaks it into the live view and the test fails.

Added the control it lacked — without an online-worker case, the test passes whenever *nothing* is returned — and the bead's second half: no test ever supplied a container with **no status**, so `c.get("status", "unknown")`, the rule that stops an unreadable container reading as running, was never exercised.

Mutation-tested: deleting the skip fails one, flipping the status default to `"running"` fails the other.

**`lu9` — the `get_payouts` mock was blind to its own filter.**

A plain `return_value` ignores kwargs, so removing `confirmed_only=True` changed nothing observable.

**Correction to the bead:** it claimed `lifetime_earned` would start counting unconfirmed payouts. It would not — `payouts.lifetime_earned` has its own `if payout.get("confirmed")` guard, so the endpoint filter is defence in depth there. My first negative control passed *because of that*, which is what surfaced it.

The field actually at risk is `confirmed_payout_count`, which has no such guard and is rendered as *"includes N confirmed payouts"* — so dropping the filter would tell the user an unconfirmed balance drop was money they received. That's what the test now asserts, and dropping the filter fails it.

**Verification:** 2501 tests, ruff clean. Every claim mutation-tested rather than asserted.

### DIFF
--- a/tests/test_main_routes.py
+++ b/tests/test_main_routes.py
@@ -174,12 +174,68 @@ class TestGetAllWorkerContainers:
         assert result[0]["_is_android"] is True
 
     def test_offline_workers_skipped(self):
+        """CashPilot-1xg: this fed a worker with NO containers.
+
+        Removing the `status != "online"` skip therefore still produced [] and
+        the test passed — it asserted nothing about the skip it is named for.
+        The offline worker now HAS a container, so deleting the skip leaks it
+        into the result and the test fails.
+        """
         from app.main import _get_all_worker_containers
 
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #192, opened 2026-08-04
fix: preflight applies the documented vps_ip default (beads batch 10)

GeiserX:
**`0yf` — one catalog fact, two readers, opposite answers.**

`services/_schema.yml` documents that `vps_ip` defaults to the opposite of `residential_ip`.

- `scripts/generate_readme_tables.py` **applied** the default → the catalog page tells the user "VPS not allowed".
- `app/preflight.py` tested `reqs.get("vps_ip") is False` — the **literal boolean** — so an absent key never entered the branch.

Measured consequence:

```
services warning today:                   5
services that should warn (with default): 26
```

Twenty-one services the docs describe as residential-only were deployed onto hosting workers with **no warning at all** — where they earn nothing or get the account banned — while the catalog page the user had just read said otherwise.

Both now read through `app.catalog.vps_allowed`, and the README generator delegates instead of keeping a second implementation. Regenerating `README.md` after the refactor produces a **byte-identical** file.

An absent `residential_ip` still yields `None` — unknown, not permitted. Returning `True` there would invent a licence the catalog never granted.

**One correction to my own test.** I first asserted that declaring both `residential_ip` and `vps_ip` is a contradiction. It isn't — an *explicit* `vps_ip` is the catalog author stating the provider accepts a VPS, and that outranks any inference. The schema defines only the *default* in terms of `residential_ip`, not exclusivity. `earnfm` legitimately declares both, so the set is pinned instead, and the combination arising by accident still surfaces.

**Verification:** 2508 tests, 95%+ coverage, ruff clean. Negative control: restoring the literal-`False` test fails.

### DIFF
--- a/app/catalog.py
+++ b/app/catalog.py
@@ -197,3 +197,24 @@ def critical_volume_targets(slug: str) -> dict[str, str] | None:
         if target:
             targets[str(target)] = str(entry.get("holds") or "Irreplaceable service state.")
     return targets
+
+
+def vps_allowed(requirements: dict[str, Any] | None) -> bool | None:
+    """Whether a VPS is permitted, applying the schema's documented default.
+
+    ``services/_schema.yml`` states that ``vps_ip`` defaults to the opposite of
+    ``residential_ip``. Two consumers read this one fact and disagreed:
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #193, opened 2026-08-04
fix: a wizard that congratulated failed deploys, and a modal missing its own recommendation (beads batch 11)

GeiserX:
Two screens that told the user something untrue.

**`4og` — the wizard congratulated you for deployments that never happened.**

The final screen was unconditional markup: *"You're all set! Your services are being deployed."* Shown whether five deployed, none did, or every deploy returned 403. `wizardState.deployed` was written twice and **read nowhere**, so a user whose deploys all failed was congratulated and sent to an empty dashboard with no idea anything had gone wrong.

Now rendered from what actually deployed, exercised in a real DOM across all four cases:

```
all deployed   -> You're all set!       Deployed: honeygain, earnapp. …
partial        -> Partly deployed       Deployed: honeygain. 1 service could not be deployed …
none deployed  -> Nothing was deployed  None of the selected services could be deployed …
nothing chosen -> Setup saved           Nothing was deployed, because no services were selected.
```

**`zr9` — the modal hid exactly the credentials the app recommends.**

Bytelixir's session cookie expires in ~2 hours; its `remember_web` and `xsrf_token` cookies last a year and are what stop collection dying the same afternoon. Both are **optional**, so the credential-health panel said *"a longer-lived credential exists for this service"* while the modal it links to offered nowhere to put it.

Storj was worse — its only field is optional, so the modal rendered **no inputs at all**.

```
storj      0 fields -> 1
bytelixir  1 field  -> 3
```

Optional fields are labelled so they aren't mistaken for required.

**Verification:** 2521 tests, 95.14% coverage, ruff clean, JS parses. Negative controls: reverting the field filter fails 1 test; reverting the wizard copy and its render call fails 2 more.

## Summary by CodeRabbit

* **New Features**
  * Credential setup dialogs now display all available fields, including clearly labeled optional inputs.
  * Setup completion now summarizes deployment outcomes, distinguishing saved, failed, partial, and complete results.
  * Completion messaging now directs users to monitor status and earnings.

* **Bug Fixes**
  * Corrected misleading deployment-success messaging.
  * Restored missing optional credential fields in setup dialogs.

### DIFF
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -1147,8 +1147,20 @@ const CP = (() => {
       if (title) title.textContent = `${col.name} — Credentials`;
 
       const secrets = config._secrets || {};
-      const fieldsHtml = col.fields.filter(f => f.required).map(f => {
+      // EVERY field, not just the required ones.
+      //
+      // Filtering to required hid exactly the credentials this UI recommends.
+      // Bytelixir's session cookie expires in about two hours; its remember_web
+      // and xsrf_token cookies last a year and are what stop collection dying
+      // the same afternoon — and both are optional, so the credential-health
+      // panel said "a longer-lived credential exists for this service" while
+      // the modal it links to offered nowhere to put it.
+      //
+      // Storj is worse: its only field is optional, so the modal rendered no
+      // inputs at all.
+      const fieldsHtml = col.fields.map(f => {
         const inputType = f.secret ? 'password' : 'text';
+        const optionalSuffix = f.required ? '' : ' <span style="color:var(--text-muted); font-weight:400;">(optional)</span>';
         // Secret inputs are write-only: empty, with a placeholder that reflects
         // whether a value is already stored (per _secrets). Non-secret fields keep
         // their plain placeholder.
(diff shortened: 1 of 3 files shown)


GeiserX:
Fixed — you were right, and it is the same kind of wrongness as the bug this PR set out to fix.

Step 3 offers **Skip to Summary**, so a user can select services and reach step 4 having deployed nothing. Keying only on `deployed.length === 0` told them *"None of the selected services could be deployed"* — sending them to hunt a failure that never happened.

The distinction is whether a deploy was **attempted**, so `deployService` records that now. Five outcomes, exercised in a real DOM:

```
all deployed     You're all set!
partial          Partly deployed (names how many did not)
attempted, 0 ok  Nothing was deployed
skipped step 3   Setup saved — nothing deployed YET
none selected    Setup saved
```

I also asserted the flag is initialised in **both** the declaration and the wizard reset — a flag nothing resets would carry the first run's state into a second pass.

Negative control: removing the skip branch fails the test that names it. 2524 tests, ruff clean.

Issue #188, opened 2026-08-04
[Bug]: Registration failed. Please try again.


radnet001:
### Description

I got this error when registering my new account

### Steps to Reproduce

1. Registration Page

### Expected Behavior

Would be able to create a new account

### Deployment Method

Docker Compose

### CashPilot Version

v1.4

### Architecture

amd64 (x86_64)

### Relevant Logs

```shell

```

### Checks

- [x] I have checked whether this is already fixed in the latest version
- [x] I searched existing issues and this is not a duplicate

GeiserX:
Thanks for reporting this — you've hit a real bug, not something you did wrong, and it's already fixed. Sorry you ran into it.

**What's happening**

Creating the first account requires a one-time **setup token**, printed to the container logs on first start. The page you landed on (`/onboarding`) never asked for it and never sent it, so the server refused with a 403 and the page showed the generic *"Registration failed. Please try again."* — which is the least helpful thing it could have said, because retrying can't work.

I confirmed this against the exact version you're running:

```
v1.4.4  onboarding.html … setup_token field: absent
v1.4.4  deps.py … first-run token gate:      present
```

The gate was added; the page it gates was never updated.

**Two ways forward — the first works right now, no upgrade**

1. **Use `/register` instead of `/onboarding`.** That page *does* have the token field, even on 1.4. Get your token with:

   ```bash
   docker compose logs cashpilot-ui | grep -i "setup token"
   ```

   You'll see a line beginning *"FIRST-RUN SETUP: no account exists yet…"* ending in the token. Go to `http://<your-host>:8080/register`, fill in the form, paste the token, and it will work.

2. **Upgrade.** The onboarding page now has the field, marks it required, tells you where to find it, and a 403 explains the actual cause instead of suggesting a retry.

**This one is on us, twice over.** Our `docker-compose.yml` pins `1.4` while its own header claims it tracks the newest release — so following the quickstart hands you a version many releases behind. I'm fixing that pin now so nobody else lands where you did.

**Could you tell me one thing?** Whichever route you take, I'd like to know it actually worked for you:

- If you tried `/register` with the token — did the account get created?
- If you upgraded — did `/onboarding` work directly?

And if it still fails, this is the most useful thing you can paste (it contains no secrets — the token is redacted):

```bash
docker compose logs --tail=50 cashpilot-ui | sed -E 's/(setup token[^:]*: )[A-Za-z0-9_-]+/\1<redacted>/i'
```

[...]

PR #194, opened 2026-08-04
fix: the quickstart installed 1.4 (issue #188), and bump cryptography past a CVE

GeiserX:
**Issue #188** is someone unable to create an account on v1.4. They got that version from us.

`docker-compose.yml` pinned `1.4` while its own header claimed the images *"track the `latest` tag"* and told the reader to replace `:latest` — **a string the file did not contain anywhere**. Anyone checking whether they were current concluded they were.

The version that hands them:

```
v1.4.4  deps.py … first-run token gate:      present
v1.4.4  onboarding.html … setup_token field: absent
```

So the owner account cannot be created at all. Fixed in v1.11.0, months ago.

**What changed**

- Both compose files pin `1.11`.
- The header now describes what the file actually does.
- A test asserts the pin equals the **newest released major.minor**, not a fixed number — so it fails the next time it falls behind, instead of quietly aging seven minor versions.
- `cryptography>=50.0.0` for the open Dependabot alert (high — PKCS#7 EnvelopedData Bleichenbacher oracle). We use Fernet, not PKCS#7 EnvelopedData, so the vulnerable path isn't one we call, but the constraint permitted the affected range.

**A verification failure worth flagging.** I first concluded v1.4 had *no* token gate — from `git show v1.4:… | grep -c` returning `0`. The tag is `v1.4.4`; `git show v1.4:` produced **nothing**, and a count of zero on empty input is indistinguishable from a confirmed absence. The corrected check prints byte counts alongside, so the two can be told apart. Had I not rechecked, I'd have told the reporter their problem was something else entirely.

**Verification:** 2514 tests, 95.14% coverage, ruff clean. Negative control: restoring the `1.4` pin fails the drift test.

## Summary by CodeRabbit

- **Improvements**
  - Updated the application and worker services to the latest 1.11 release series.
  - Improved release consistency by pinning service images to major and minor versions while allowing patch-level updates.
  - Clarified deployment guidance so image version behavior is accurately documented.
  - Strengthened security and compatibility support through updated encryption components.


GeiserX:
All three addressed — the first one mattered most.

**The CVE fix was incomplete.** Dependabot names `uv.lock` as the manifest, and the lock still declared `>=48.0.1` and pinned `cryptography 48.0.1` — inside the vulnerable range. Changing `pyproject.toml` and `requirements.txt` alone would have left the alert open while looking fixed. Regenerated with `uv lock --upgrade-package cryptography`; now 50.0.0, and the diff touches nothing else.

**The drift test would have skipped in CI.** `actions/checkout` doesn't fetch tags, so `git tag` returns nothing and the test hit its own `pytest.skip` — decorative in the one place it's meant to run, which is the same shape of defect as the stale pin it guards. CI now fetches tags, and the test **fails rather than skips** when CI has none, so a workflow regression can't quietly disarm it.

**The commented template is checked now.** `docker-compose.fleet.yml`'s commented remote-worker block is something users uncomment and run, so a stale pin there ships an old image just as surely as a live one — and my first version stripped comments *specifically* to ignore it. Negative control: planting `1.4` there now fails the test.

One thing I'll state plainly rather than imply: this venv has no `pip`, so my local run exercised 48.0.0. I verified 50.0.0 exists and installs cleanly, and CI installs from `requirements.txt`, so CI is what actually runs the patched library.

PR #195, opened 2026-08-04
fix: an undocumented payout minimum is not zero, and settings that never applied (beads batch 12)

GeiserX:
Two claims the app could not support.

**`x26` — 21 services encoded "we don't know" as "no minimum".**

`payouts.min_payout` is explicitly three-valued: positive is a real minimum, `0.0` is a documented *"cash out any amount"*, `None` is undocumented. Twenty-one services carried `min_amount: 0` with an **empty** `payment.minimum_payout` — the catalog asserting a fact nobody had established.

Two user-facing consequences:

- `api_earnings_breakdown` reported the user **eligible to cash out any balance above zero**, sending them to a withdrawal page that would refuse them.
- `looks_like_payout` returns `False` on a falsy threshold, so **payout detection was silently disabled** for all 21.

Those minimums are `null` now, and eligibility is three-valued.

**The existing tests were right and my first attempt was wrong.** I collapsed *no cashout section* and *unknown minimum* both to `None`, which broke two tests in `test_eligibility.py`. They were protecting a real distinction: no cashout section is a definite **`False`** — there is no withdrawal route at all — while a route with no published minimum is genuinely **unknown**.

**`73o` — Settings asserted a stored value was in force.**

Every variable the page offers is resolved once from the environment at import, so a value written to the config table can never take effect. The UI showed a green *"Variables saved"* toast and a **"DB"** badge meaning *the database value is in use*, while collection carried on at the old interval.

Fixed by telling the truth rather than wiring them up, because several **cannot** change at runtime: the session key signs logins already in use, the data directory is where the open database lives, and the fleet key is what enrolled workers already hold. A stored value now reads **"Stored, not applied"**, and the page says to set it in the compose file and restart.

**Verification:** 2542 tests, 95.11% coverage, ruff clean (both `check` and `format --check`), JS parses. Negative controls: collapsing eligibility fails one test; restoring the DB badge fails another.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1880,7 +1880,16 @@ async def api_earnings_breakdown(request: Request) -> list[dict[str, Any]]:
         slug = row["platform"]
         svc = catalog.get_service(slug)
         cashout = (svc.get("cashout", {}) if svc else {}) or {}
-        min_amount = float(cashout.get("min_amount", 0))
+        # Three-valued, matching payouts.min_payout: a positive number is a
+        # real minimum, 0 is a documented "no minimum", and None means nobody
+        # published one. float(..., 0) collapsed the last two, so 21 services
+        # with no published minimum reported the user as eligible to cash out
+        # any balance above zero.
+        raw_min = cashout.get("min_amount")
+        try:
+            min_amount = float(raw_min) if raw_min is not None else None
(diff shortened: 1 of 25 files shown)


GeiserX:
@coderabbitai review

PR #196, opened 2026-08-04
fix: net earnings subtracted a EUR cost from a USD gross (beads batch 13)

GeiserX:
**`dlr`** — `gross` is USD by contract; `cost` comes from a tariff in `power_currency`. Subtracting one from the other produced a number in neither, labelled with the tariff currency.

At ~1.08 USD/EUR that's an **8% error, in the direction that flatters the result**, on the single figure that decides whether a machine is worth keeping powered on.

**The tariff is converted into USD, not the gross into the tariff currency.** My first attempt did the latter and it was wrong: every other money figure in this API is canonical USD and the frontend has a display-currency layer, so converting the gross would have made this endpoint the exception. Converting the price keeps the convention and needs no new rate direction.

With no rate available it reports `cost_known: false` and says why, rather than inventing a net. That mirrors what the module already does with no tariff at all — *"a zero cost would render gross as net and quietly overstate earnings."* Passing `"USD"` as a label while the price stayed EUR-denominated (my second attempt) only **moves** the mixing.

**`conftest` now seeds the fiat rates a running system always has.** `exchange_rates.refresh()` populates them at startup and every 15 minutes, but the test process never calls it, so `_fiat_rates` was empty everywhere. Ten tests configure a EUR tariff and expect a cost; with no rate the endpoint correctly refuses, so they failed for reasons unrelated to what they test.

Two existing assertions changed, both encoding the defect — `test_power` expected `currency == "EUR"` on a figure whose gross was USD.

**Verification:** 2533 tests, 95.12% coverage, both ruff gates, JS parses. Negative control: restoring the mixing fails four tests.

## Summary by CodeRabbit

* **New Features**
  * Net earnings now support electricity tariffs priced in non-USD currencies by converting them to USD.
  * Responses identify the tariff currency and explain when conversion rates are unavailable.
* **Bug Fixes**
  * Prevented misleading net earnings when currency conversion data is missing.
  * Corrected exchange-rate direction and preserved gross earnings in USD.
  * Improved handling of native-currency earnings and pricing scenarios.
* **Tests**
  * Added coverage for currency transitions, tariff conversion, missing rates, and related earnings calculations.

### DIFF
--- a/app/exchange_rates.py
+++ b/app/exchange_rates.py
@@ -160,6 +160,25 @@ def get_all() -> dict[str, Any]:
     }
 
 
+def from_usd(amount: float, currency: str) -> float | None:
+    """Convert a USD amount into *currency*. The inverse of :func:`to_usd`.
+
+    ``_fiat_rates`` stores USD->X rates, so this MULTIPLIES where ``to_usd``
+    divides. Getting that backwards would be worse than not converting at all,
+    which is why the two live next to each other.
+
+    Fiat only: a USD figure has no meaningful expression in a provider's token,
+    and inventing one would put a crypto amount where a currency belongs.
+    Returns None when no rate is available.
+    """
+    if currency == "USD":
+        return amount
+    rate = _fiat_rates.get(currency)
+    if rate and rate > 0:
+        return amount * rate
+    return None
+
+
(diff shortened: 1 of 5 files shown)


GeiserX:
@coderabbitai review

PR #197, opened 2026-08-04
fix(worker): removing a worker no longer locks the host out permanently

GeiserX:
## The defect

Removing a worker in the fleet dashboard deletes its row and its per-worker key on the UI side. The worker never learned that.

`app/worker_api.py` `_active_key()` returns `_worker_key or API_KEY` — once `/data/.worker_key` exists, the shared key is **never sent again**. So the host 401s forever, while its service containers keep running and keep earning. Nothing on either side explains it: the UI shows one fewer worker, the worker logs a bare 401.

Both places that described this promised a recovery no code performed:

- `docs/upgrade-v1.md:51` — removing the worker "(clears its enrollment) so the shared key is accepted again". Nothing cleared anything.
- `app/templates/fleet.html:346` — `Remove worker "${name}"? This will unregister it from the fleet.`

The only real recovery was SSHing in to delete a file documented nowhere.

## The fix

After **ten consecutive** rejections of our own key, the worker discards it and re-enrols with the shared key.

Bounded deliberately. Discarding on the first failure would re-enrol on any transient blip and widen the window in which the shared key is accepted — the exact weakening per-worker keys exist to close. The threshold sits above the existing `_AUTH_FAILURE_ALARM_AFTER = 3`, so the operator is told before anything is deleted. A key that cannot be removed from disk is **kept**, not zeroed in memory: believing we re-enrolled while the stale key survives a restart brings the same lockout back with no failures leading up to it.

The docs and the confirm dialog now describe what actually happens, including the manual shortcut.

## Evidence

- `ruff check .` + `ruff format --check .` clean; `node --check` on all JS
- 2558 tests pass
- **Negative control:** removing the five-line discard branch (module still imports — a behaviour change, not a syntax break) fails `test_sustained_rejection_discards_the_stale_key` and `test_it_says_why_in_the_log`, and nothing else. Reverting the two doc changes fails their two tests.
- Controls that keep the fix bounded: 9 consecutive 401s discard nothing; a flaky link (`401, 500` alternating) discards nothing; a success before the threshold saves the key.

Closes CashPilot-u10

### DIFF
--- a/app/templates/fleet.html
+++ b/app/templates/fleet.html
@@ -343,7 +343,7 @@ CASHPILOT_WORKER_NAME=server-name</code>
   };
 
   async function removeWorker(workerId, name) {
-    if (!confirm(`Remove worker "${name}"? This will unregister it from the fleet.`)) return;
+    if (!confirm(`Remove worker "${name}"?\n\nIt is removed from the fleet here, but the worker itself keeps running. It will re-enrol on its own after a few minutes of rejected heartbeats. To bring it back immediately, delete /data/.worker_key on that host and restart the container.`)) return;
     try {
       await CP.api(`/api/workers/${workerId}`, { method: 'DELETE' });
       CP.toast(`Worker "${name}" removed`, 'success');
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #198, opened 2026-08-04
test: prove the auth guard for every route instead of 14 hand-listed handlers

GeiserX:
## The defect

`tests/test_audit_guards.py` checks **14 handler names typed out by hand**. `test_main_routes.py` adds one more. The app registers **75 routes**.

A guard deleted from any of the other 61 kept the whole suite green. Verified by removing:

- `_require_auth_api` from `api_exchange_rates`
- `_require_writer` from `api_collect`
- `_require_auth_api` from `api_earnings_flatlines`

2300 passed each time. An anonymous `GET /api/exchange-rates` against the mutated tree then returned the full rate table. Every endpoint test patches the guard away before calling the handler, so nothing anywhere notices when the real one is gone.

## The fix

Stop maintaining a list. This drives the real ASGI app with no session and no key, and requires **every registered route** to refuse — so a new endpoint is covered the moment it exists, and an author who wants one public has to come here and say why.

Two things were needed to make that mean anything:

- **Eleven POST routes answered 422 to an empty body.** FastAPI validates before the handler runs, so the guard was never reached and the 422 proved nothing. Bodies are now built from each route's own Pydantic model. A separate test asserts no route is ever accepted on a 422.
- **`/api/workers/heartbeat` answered 503** "Fleet key not configured" — a fail-closed refusal, not a guard. Accepting it would have left the one route a stolen shared key targets effectively untested.

An earlier attempt inspected each handler's AST for a guard call and reported 17 unguarded routes, **all false**: `api_stop` delegates to `_svc_stop`, which guards, and the page routes redirect to `/login`. Driving the app sees both; reading the handler cannot.

## Evidence

**Full mutation sweep.** Each of the 55 `_require_*(request)` call sites in `main.py` was deleted in turn and this file re-run:

| outcome | count |
|---|---|
| turns this file red | 52 |
| leaves a module that will not import (empty `if`/`else` body) | 3 |
| **survives** | **0** |

For the three endpoints named in the bead, the old allowlist test stayed green (40 passed) while this one went red — both halves proven.

`ruff check .` + `ruff format --check .` clean. 2620 passed, 6 skipped.

## Stated boundary

A guard that only fires on a **branch** — `_require_owner` inside `if allow_delete_critical`, the deploy arm of `api_worker_command` — is unreachable for an anonymous caller, who is refused by the outer `_require_writer` first. Role separation is a different question from authentication and stays with `TestMutatingEndpointsSitAboveViewer`. A test here fails if that class is ever deleted, so the division of labour cannot silently become a hole.

Closes CashPilot-zfd

### DIFF
--- a/tests/test_audit_guards.py
+++ b/tests/test_audit_guards.py
@@ -40,6 +40,12 @@ class TestEveryEndpointRejectsAnAnonymousCaller:
 
     Every other endpoint test patches `_require_auth_api` to a no-op before
     calling the handler, so nothing anywhere noticed the guard was gone.
+
+    The class name overstates what a hand-written list can do — it covered 14 of
+    75 routes, and a guard deleted from any of the other 61 kept the suite green.
+    tests/test_every_route_rejects_anonymous.py now sweeps every registered route
+    instead. This is kept because it calls the handlers directly, which catches a
+    guard that is present but unreachable behind an earlier return.
     """
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #200, opened 2026-08-04
fix: a dollar balance was compared against a token minimum at 1:1

GeiserX:
## The defect

The balance is recorded in whatever **the collector** reports. The payout minimum is declared in whatever **the provider** cashes out in. Thirteen of the fifteen registered collectors agree — which is why nothing noticed. Two do not:

| service | collector reports | catalog declares |
|---|---|---|
| storj | `USD` (its API gives cents) | `4.0 STORJ` |
| anyone-protocol | `USD` | `ANYONE` |

So a real **$3.50** Storj balance rendered as `Balance now: 3.50 STORJ`, the card said **"0.50 to go"** toward a threshold in a different unit, and the service-row tooltip printed 4 STORJ as **"$4.00"**. Every number the user was asked to compare was in a unit it was not in.

The same 1:1 comparison decided `eligible`, so a user could be told they can cash out and sent to a page that refuses them.

## The fix

Reconciled **once**, in `payouts.min_payout_in`, and used everywhere the two meet:

- eligibility (`/api/earnings/breakdown`)
- the projection countdown (`payouts.project`)
- the payout-progress card
- the service row + its tooltip
- the claim modal
- the "closest to payout" sort

`/api/services/{slug}/payout-progress` now also states what unit its figures are in, instead of leaving the frontend to assume the catalog's.

Three-valued throughout, matching the rest of the codebase:

- **no rate → `None`**, no comparison and no estimate. Unknown is not zero and not "eligible".
- **a documented `0` minimum needs no rate** — zero of anything is zero; refusing there would hide a real answer.
- **an undeclared cashout currency is taken at face value** — correct whenever the units match, which is almost every catalog entry. Refusing there would disable the payout bar catalog-wide, a far worse answer than the behaviour being fixed.

`payouts.project` gained the balance's currency as a **defaulted** argument, so no existing caller changes behaviour.

## Evidence

`ruff check` + `ruff format --check` clean, `node --check` and the currency harness clean, **2644 passed**.

Negative control — five parts, no overlap:

|---|---|
| `min_payout_in` never converts (the original behaviour) | 7 |
| eligibility compares against the declared minimum | 2 |
| the projection counts down to the declared minimum | 2 |
| the card labels with the cashout unit | 1 |
| the rows divide by the declared minimum | 2 |

Two tests also pin the catalog case itself (Storj still declares STORJ, its collector still reports USD), so a later YAML tidy-up cannot quietly stop exercising this.

Closes CashPilot-c50

## Summary by CodeRabbit

* **Bug Fixes**
  * Payout eligibility and progress now correctly compare thresholds across different currencies.
  * Added currency-aware payout projections, remaining-balance calculations, and countdowns.
  * Unconvertible currency mismatches no longer produce misleading threshold estimates.
  * Updated payout sorting and eligibility displays to use comparable amounts.

* **Tests**
  * Added coverage for cross-currency payout thresholds and projections.
  * Improved resilience of projection-related validation.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2515,7 +2515,11 @@ async def api_payout_progress(request: Request, slug: str) -> dict[str, Any]:
         # What the numbers above are actually in, so nothing has to assume.
         "balance_currency": balance_currency,
         "min_amount": minimum,
-        "min_amount_currency": balance_currency if minimum is not None else declared_currency,
+        # Falls back to the DECLARED unit when nothing has been collected: with
+        # no balance currency, min_payout_in returns the catalog figure at face
+        # value, and labelling that with None leaves the consumer holding a
+        # threshold it cannot attribute to any unit.
+        "min_amount_currency": (balance_currency if (minimum is not None and balance_currency) else declared_currency),
         # False only when a real minimum exists and could not be brought into
         # the balance's unit. The card hides the comparison rather than drawing
         # a bar out of two different currencies.
--- a/app/main.py
+++ b/app/main.py
@@ -1891,6 +1891,12 @@ async def api_earnings_breakdown(request: Request) -> list[dict[str, Any]]:
         except (TypeError, ValueError):
             min_amount = None
         balance = float(row["balance"])
+        # The minimum in the same unit as this balance. The catalog declares it
+        # in whatever the provider cashes out in, which for Storj (USD balance,
+        # STORJ minimum) and anyone-protocol is not what the collector reports —
+        # so `balance >= min_amount` was comparing two different currencies at
+        # 1:1, and told the user they were eligible when they were not.
+        comparable_min = payouts.min_payout_in(svc, row.get("currency"))
         prev_balance = float(row.get("prev_balance", 0))
         delta = balance - prev_balance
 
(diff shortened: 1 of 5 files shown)


GeiserX:
Addressed all three findings — each was real.

1. **Token conversion** — you were right that the test was testing the mock. `from_usd` is deliberately fiat-only, so every token target returned `None` in production while the lambda made it look fine. Rather than widen `from_usd`'s contract, the inverse is now derived from `to_usd` itself (one unit of the target is worth `to_usd(1.0, target)`), which handles fiat and crypto alike and cannot be inverted by mistake. The test drives the real rate table now, plus an unpriced-token control.

2. **Threshold unit** — applied your fallback to `declared_currency`.

3. **`eligible: null`** — the sharpest of the three, and it was my own regression: the row coerced null to 0 and then rated every positive balance eligible, while the modal called the same state "Below minimum payout". The row now takes eligibility from the endpoint instead of recomputing it, and the modal has an explicit "Cannot tell yet" state naming the two currencies that could not be reconciled.

Each fix has a negative control that fails exactly one test when reverted.

PR #207, opened 2026-08-04
fix: three catalog entries stated their payout minimum in two units

GeiserX:
## The defect

`services/_schema.yml` defines `cashout.min_amount` as *"numeric, derived from `payment.minimum_payout`"*. So `cashout.currency` is **the unit that number is in** — not the token the provider pays in, which `payment.currency` already records.

Three entries disagreed with themselves:

| slug | `payment.minimum_payout` | `cashout.currency` |
|---|---|---|
| storj | `"$4"` | `STORJ` |
| ebesucher | `"$2"` | `EUR` |
| proxybase-xyz | `"$1"` | `USDC` |

The bead named only storj; the other two came from checking the same rule across the whole catalog.

For storj it's a **real distortion**: 4 STORJ is around a dollar, so the user was told they needed roughly four times what the provider actually asks. #200 fixed the *arithmetic*, but it reconciles against what the catalog **declares** — a wrong declaration still produces a wrong threshold. This fixes the declaration.

`payment.currency` is untouched in all three: STORJ, EUR and USDC are what the provider actually pays in, and that's worth keeping.

A catalog-integrity test now fails on any entry documenting its minimum in dollars while labelling the derived number differently.

## A test I had to correct, not just add

#200's guard asserted *"storj still declares a token minimum"* — pinning that fix to the very declaration this bead shows is **wrong**. A test standing in the way of fixing the fact it pinned.

It now asserts what it was actually for: that a mismatch between the balance's unit and the declared minimum is still *reconciled* rather than compared at 1:1 — checked against the function directly, so it holds however the catalog is written.

## Evidence

`ruff check` + `ruff format --check` clean, **95.22% coverage**, 2742 passed.

Negative control: restoring storj's `STORJ` declaration fails 4 tests. The revert prints its byte count, because a control that silently fails to apply is worse than none.

## One open factual question

**ebesucher** documents `minimum_payout: "$2"` while paying in **EUR**. Making `cashout.currency` USD is the internally consistent reading (the number was derived from a dollar string), and it's what the schema requires — but whether the provider's real threshold is $2 or €2 is a fact about ebesucher that the repo can't settle. Flagging rather than asserting; the amounts are close enough that it doesn't change behaviour materially today.

Closes CashPilot-bgj

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected cashout currency labels for several services to consistently use USD.
  * Preserved provider payout currencies while ensuring documented cashout minimums are interpreted in dollars.
  * Corrected Storj’s reconciled cashout threshold to remain $4 without unnecessary token conversion.

* **Tests**
  * Added coverage validating currency labels, payout-token preservation, thresholds, and exchange-rate independence.

### DIFF
--- a/services/bandwidth/ebesucher.yml
+++ b/services/bandwidth/ebesucher.yml
@@ -47,7 +47,7 @@ cashout:
   method: redirect
   dashboard_url: "https://www.ebesucher.com/account"
   min_amount: 2.0
-  currency: EUR
+  currency: USD
   notes: "Euro payout via PayPal or bank transfer."
 
 platforms: [browser-extension]
--- a/services/bandwidth/proxybase-xyz.yml
+++ b/services/bandwidth/proxybase-xyz.yml
@@ -70,7 +70,7 @@ cashout:
   method: manual
   dashboard_url: "https://proxybase.xyz"
   min_amount: 1.0
-  currency: USDC
+  currency: USD
   notes: "Lock earnings via <code>proxybase-cli seller payout create</code>, received as USDC on Tempo Chain (Tempo MPP). Requires a Tempo wallet address."
 
 platforms: [docker, linux, macos, windows]
--- a/services/storage/storj.yml
+++ b/services/storage/storj.yml
@@ -109,7 +109,7 @@ cashout:
   method: redirect
   dashboard_url: "http://localhost:14002"
   min_amount: 4.0
-  currency: STORJ
+  currency: USD
   notes: "Payouts sent monthly in STORJ tokens to configured wallet address. Dashboard at port 14002."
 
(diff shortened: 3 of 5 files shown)


GeiserX:
Applied — you were right, and it was the sharper kind of wrong: the guard's docstring claimed to catch a flattened `cashout.currency`, but the predicate read `payment.currency`, which this change never touches. It would have held even in the exact scenario it exists to catch.

Now reads `cashout.currency`, plus a floor of ten (21 services still declare a token minimum) so one stray edit can't satisfy it either. Controlled by flattening `cashout.currency` across all 21 files — fails now, passed before.

PR #212, opened 2026-08-04
fix: a container you started yourself got live buttons that 404

GeiserX:
## The defect

The worker **already answers this**. `app/orchestrator.py` sets `"deployed_by": "external"` for a container matched by **image** rather than by CashPilot's own label — one the user started themselves with `docker run`.

Then `_get_all_worker_containers` rebuilt each container dict from scratch:

```python
"deployed_by": worker_name,     # ← the worker's answer, discarded
```

`grep -rn deployed_by app/` showed the only readers were the two writers in `orchestrator.py` and these two overwrites.

So a container you started yourself appeared as an ordinary managed service — "Running" badge, instance count, CPU/memory, and enabled **Restart / Stop / Logs**. Clicking any of them returned:

> 404 Container for honeygain not found

…which reads as a CashPilot bug on a service the same screen had just called Running.

## The fix

The worker's answer survives. The node name was never at risk — `_node` already carries it, and that's what the UI labels instances with.

**A row is flagged only when *every* instance was started outside CashPilot.** With a mix, the managed one can still be controlled, so the row keeps its buttons and the per-instance flag marks the odd one out. Flagging the whole row would remove the buttons that *do* work.

The UI disables the three container actions and **says why**. A dead button with no explanation reads as a broken app — the same complaint as the 404, just quieter. The pre-existing "No Docker access" reason is preserved, pinned by a control.

## Evidence

`ruff check` + `ruff format --check` clean, `node --check` and both CI harnesses clean, **95.26% coverage**, 2825 passed.

Negative control:

|---|---|
| overwrite `deployed_by` again (the original bug) | 1 |
| flag the row on *any* external instance rather than all | 1 |
| leave the buttons enabled | 2 |

Controls also pin that a managed container is unaffected, an older worker sending no `deployed_by` still falls back to the node name, and the worker still computes the flag at all — without that last one, everything here is inert.

Closes CashPilot-ljx

## Summary by CodeRabbit

* **New Features**
  * Service listings now identify containers started outside CashPilot with an “Unmanaged” badge.
  * Deployment status now reports unmanaged state for individual instances and entire services.
* **Improvements**
  * Controls for restarting, stopping, or viewing logs are disabled for unmanaged single-instance services, with explanatory guidance.
  * Container deployment classifications are preserved more accurately, including Android app deployments.
* **Bug Fixes**
  * Improved detection and reporting of externally deployed containers across workers and services.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -204,7 +204,16 @@ async def _get_all_worker_containers(workers: list[dict[str, Any]] | None = None
                             "net_rx_bytes": c.get("net_rx_bytes"),
                             "net_tx_bytes": c.get("net_tx_bytes"),
                             "category": "",
-                            "deployed_by": worker_name,
+                            # The WORKER's answer, not the node name. The image
+                            # matcher sets "external" for a container it found by
+                            # image rather than by CashPilot's own label — one the
+                            # user started themselves. Overwriting it here meant
+                            # that container appeared as an ordinary managed
+                            # service with live Restart/Stop/Logs buttons, and
+                            # every one of them answered "404 Container not
+                            # found" for a row the same screen called Running.
+                            # The node name is already carried by _node.
+                            "deployed_by": c.get("deployed_by") or worker_name,
                             "_node": worker_name,
                             "_worker_id": w.get("id"),
                             "_has_docker": worker_has_docker,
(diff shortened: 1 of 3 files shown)


GeiserX:
Fixed — and this was the sharper half of the change, not a detail.

The mixed case is the one I deliberately designed for: the row keeps its buttons because the managed instance can still be controlled, and the per-instance flag was supposed to mark the odd one out. **The sub-row renderer never read that flag.** So in exactly the scenario the PR description described as handled, the external instance still offered Restart/Stop/Logs and every one still 404'd.

Two of my new tests were also weaker than they looked: they asserted `iUnmanaged` was **declared**, so a revert deleting only the `disabledAttr` ternary left the variable unused and both still passed against the broken state. They now assert every `disabledAttr` assignment carries the reason — the use, not the declaration.

Control: reverting the sub-row ternary now fails 3 tests; before strengthening it failed 1.

PR #220, opened 2026-08-04
fix: two worker rows, one name, and no way to tell them apart

GeiserX:
## The defect

A worker redeployed with a new or empty `/data` — appdata path changed, volume recreated — registers under a **fresh random `client_id`**. The old row stays in the fleet list forever: offline, with its last-known container pills, inflating the worker count.

Both rows carry the **same display name**, and the page rendered **no id at all**. So the operator picks one to Remove by guesswork — and removing the live one takes that host out of the fleet.

## Why the automatic purge isn't the answer

It only deletes rows with no `api_key_enc`, and **the first heartbeat of every worker sets that** — so no row that ever heartbeated is eligible. `_check_stale_workers` records the reason: a host persists its key locally and re-presents it, so deleting an enrolled row leaves no match for that key.

Widening the purge would trade a cosmetic duplicate for an outage. A test now pins that it stays as it is, with the reasoning intact.

## The fix

What was missing is the information needed to choose correctly:

- every card shows the **`client_id`** (falling back to `#rowid` for pre-`client_id` workers)
- a **duplicated name is flagged**, with the cause and what to compare
- the **removal prompt names the id**

## Evidence

`ruff check` + `ruff format --check` clean, both CI harnesses clean, **95.26% coverage**, 2963 passed.

Negative control:

|---|---|
| remove the id element (the original state) | 4 |
| drop the duplicate flag | 3 |
| take the id out of the removal prompt | 1 |

The id tests were **rewritten after the first control caught them**. They matched `w.client_id ||` anywhere in the file — and that expression appears six times in the per-machine power inputs from #219, which are data attributes the user never reads. Deleting the *visible* id left them passing. They now pin the rendered element.

Closes CashPilot-3fr

## Summary by CodeRabbit

- **Improvements**
  - Fleet worker cards now display each worker’s client ID, making similarly named workers easier to distinguish.
  - Duplicate worker names are clearly flagged with guidance for comparison and redeployment.
  - Worker removal confirmations now identify the specific worker being removed while retaining the existing re-enrollment warning.

- **Bug Fixes**
  - Improved worker removal safety by preventing ambiguity when multiple workers share the same display name.
  - Automatic purging behavior remains limited to unenrolled workers.

### DIFF
--- a/app/templates/fleet.html
+++ b/app/templates/fleet.html
@@ -259,6 +259,18 @@ CASHPILOT_WORKER_NAME=server-name</code>
     const container = document.getElementById('fleet-workers');
     if (!container) return;
 
+    // A worker redeployed with a new or empty /data registers under a fresh
+    // client_id, and the old row stays forever — offline, same display name,
+    // last-known container pills. The operator then picks one to Remove by
+    // guesswork, and removing the LIVE one used to lock that host out
+    // permanently. Naming the duplicates and showing the id is what makes that
+    // choice safe (CashPilot-3fr).
+    const nameCounts = {};
+    for (const w of workers) {
+      const key = String(w.name || '');
+      nameCounts[key] = (nameCounts[key] || 0) + 1;
+    }
+
     let html = '';
 
     if (workers.length === 0) {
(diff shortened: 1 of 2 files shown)


GeiserX:
Fixed in #221 — you were right, and it is reachable rather than theoretical: a worker name comes from `CASHPILOT_WORKER_NAME`.

Measured rather than reasoned about:

```
counts["__proto__"] = (counts["__proto__"] || 0) + 1   ->  {}
counts["toString"]  = (counts["toString"]  || 0) + 1   ->  "function toString() { [native code] }11"
```

So the count is not a number, `> 1` is false, and two workers with that name are never flagged — the one case the feature exists for. Affects `__proto__`, `toString`, `constructor`, `valueOf`, `hasOwnProperty`.

`Object.create(null)`, plus the counting extracted into a named function so the node harness can **run** it with all five names rather than read the source. Reverting to `{}` fails six assertions.

Also: I merged this PR before reading your comment, which was out of order on my part.

PR #232, opened 2026-08-05
fix: collect credentials that were stored before saving them began tracking

GeiserX:
## The bead

`CashPilot-h3d` — Settings → Collectors reports "Configured" for credentials that can never be used.

Half of the premise is already false. #187 made `POST /api/config` create an `external` deployment row whenever a service's required credentials are complete, so at save time the badge and the tracking agree.

The other half survived, on the path nobody exercises by hand: **a database written before that fix.**

- credentials are complete → Settings shows the green **Configured** badge
- no deployment row exists → `make_collectors` never builds a collector
- no earnings ever arrive
- nothing re-runs the save path, because the user has no reason to re-save credentials the UI says are fine

The fix that was supposed to end this state left every existing install sitting in it.

## The change

- `collectors.fully_configured_slugs(config)` — one predicate for "these credentials are complete", now shared by the save path and the backfill instead of a second inline copy that can drift.
- `_track_fully_configured_services()` runs once per start, before the collection job is scheduled, and gives every fully-credentialled service its tracking row.

Deliberately narrow:

- **Idempotent.** A slug that already has a row of any status is left exactly as it is — that row can carry a real container id and an encrypted spec, and replacing it with an empty placeholder would lose the spec a later redeploy needs.
- **Required credentials only.** Optional args are the ones a collector works without; demanding them would refuse to track a service that would collect perfectly well.
- **A collector with no required args is never claimed.** There is no credential to be complete, so "the user configured this" would assert something the user never did. (`storj` is the live case.)
- **A database failure cannot block startup** — this runs inside `lifespan`.

## Evidence

18 new tests in `tests/test_beads_batch_46.py`. One of them drives `make_collectors` over the backfilled row and asserts a Honeygain collector comes out, because a row of the wrong shape would satisfy "a row exists" and still collect nothing.

Four negative controls, all caught:

| reverted | result |
|---|---|
| backfill defined but never called from `lifespan` | 2 failed |
| backfill overwrites an existing deployment row | 3 failed |
| predicate accepts partial credentials (`all` → `any`) | 3 failed |
| save path stops scoping to the slugs the request touched | 1 failed |

Sources verified byte-identical to the originals after every revert.

Gates: `ruff check` clean, `ruff format --check` clean, `currency_check.mjs` PASS, `fleet_staleness_check.mjs` PASS (20 assertions), **3118 passed, 95.29% coverage**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Services with complete required credentials are now automatically prepared at startup, even when deployment records are missing.
  * Existing deployment records are preserved, and repeated startup processing avoids creating duplicates.
  * Credential validation is now applied consistently when saving configuration.
* **Reliability**
  * Improved handling of incomplete, optional-only, unsupported, and failed service configurations.

### DIFF
--- a/app/collectors/__init__.py
+++ b/app/collectors/__init__.py
@@ -232,6 +232,32 @@ def make_collectors(
     return collectors
 
 
+def fully_configured_slugs(config: dict[str, str]) -> set[str]:
+    """Slugs whose every REQUIRED credential key has a value in ``config``.
+
+    The one predicate behind "this service's credentials are complete". It is
+    what Settings renders as the green "Configured" badge, what decides whether
+    saving credentials starts tracking a service, and what the startup backfill
+    uses to catch credentials stored before tracking existed. Three copies of it
+    would eventually disagree, and the failure when they do is silent: a badge
+    that says Configured for a service nothing collects.
+
+    REQUIRED only. Optional args (``?``-prefixed in ``_COLLECTOR_ARGS``) are
+    exactly the ones a collector works without, so demanding them would refuse
+    to track a service that would collect perfectly well.
+
+    A collector with no required args at all is NOT included. There is no
+    credential to be complete, so "the user configured this" would be asserting
+    something the user never did.
+    """
+    configured: set[str] = set()
+    for slug in COLLECTOR_MAP:
+        required = [f"{slug}_{arg}" for arg in _COLLECTOR_ARGS.get(slug, []) if not arg.startswith("?")]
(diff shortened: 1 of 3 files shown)


GeiserX:
CodeRabbit's cache-eviction finding was real and is fixed in 9dec46c. Confirmed by driving it before the fix: a second `make_collectors` with identical config returned the same object the test had already closed.

Its follow-up review is rate-limited, so this merges on green CI + CodeQL + the four negative controls above.

PR #233, opened 2026-08-05
fix: bound the window in which the shared key still works for a worker

GeiserX:
## The bead

`CashPilot-cvr` — a worker that never confirms its key keeps the shared bootstrap key valid for its identity forever, and is handed the key again every 60 seconds.

**Confirmed against the code**, not just the report. `_authenticate_worker_heartbeat` returned `"reissue"` on `not confirmed and shared_ok` with nothing bounding how long a key may stay unconfirmed — while its own docstring called that *"a bounded window that closes on the worker's first own-key heartbeat"*.

For a worker that **cannot persist its key**, that heartbeat never arrives:

- a pre-1.0.0 worker image
- a `/data` that is read-only
- a `/data` that is not a persistent volume

The security cutover the per-worker keys exist for silently never completed. Anyone holding `CASHPILOT_API_KEY` could go on speaking as that worker indefinitely, and the UI re-sent the per-worker key to them once a minute. The only trace was one log line a minute, which nobody reads.

`docs/upgrade-v1.md` made it worse by telling users an old worker *"can no longer heartbeat"* — so nobody went looking for one that still was.

## The change

The window now also closes **24 hours after the key is minted**, on a wall clock the worker cannot influence. An attempt count was the alternative and is worse: it just re-expresses the heartbeat interval, and breaks when that interval is configured differently.

- `key_issued_at` on `workers`, stamped whenever a key is minted.
- `enrollment_state()` — pure, shared by the heartbeat guard and the fleet page, so the two cannot disagree about who is still enrolling.
- Past the window the shared key is **refused** for that worker, with a message naming the cause and the recovery.
- The fleet page marks it **enrollment incomplete**, with a tooltip explaining `/data/.worker_key` and how to fix it. There is also a quiet **enrolling** badge during the window.
- The upgrade doc now says what actually happens.

**Absent is not expired.** Two deliberate choices, both about not turning a security fix into an outage:

- A missing or unparseable issue time reads as *still enrolling*. Locking a worker out on the strength of a value nobody wrote is the wrong direction.
- The migration backfills existing unconfirmed workers to **now**, not to NULL and not to the distant past — otherwise this patch would refuse every mid-enrollment worker the moment it landed, and take a working fleet offline with no warning.

The steady state is untouched: a confirmed worker presenting its own key returns `"ok"` before this check is reached (a test asserts the issue time is never even read on that path).

## Evidence

26 new tests in `tests/test_beads_batch_47.py`, including a real pre-migration database built without the column, to prove the upgrade does not lock its worker out.

Seven negative controls, all caught:

| reverted | result |
|---|---|
| window not enforced (original behaviour) | 3 failed |
| unknown timestamp treated as expired | 2 failed |
| key issue time never stamped on mint | 1 failed |
| migration backfills NULL instead of now | 1 failed |
| state computed but not published to the fleet page | 1 failed |
| badge removed from the fleet page | 1 failed |
| upgrade doc keeps its false claim | 1 failed |

All four source files verified byte-identical to the originals after every revert.

Gates: `ruff check` clean, `ruff format --check` clean, `currency_check.mjs` PASS, `fleet_staleness_check.mjs` PASS (20 assertions), **3144 passed, 95.32% coverage**.


GeiserX:
@coderabbitai review

GeiserX:
CodeRabbit is rate-limited and declines to re-review, so this merges on green CI + CodeQL + the seven negative controls.

Adversarial pass over the auth-path edge cases before merging:

- **Worker offline >24h that DID persist its key** — presents its own key, returns `ok` before the window check. Unaffected.
- **Enrollment response dropped AND worker down >24h** — the one case that now needs operator action. It is visible (`enrollment incomplete`) and the 401 says what to do.
- **Stale-worker purge** — such a worker has `api_key_enc` set, so it is never auto-deleted; it stays on the fleet page with the badge rather than vanishing.
- **UI restored from backup** with `key_confirmed=0` and an old issue time — the worker still holds its key, presents it, and confirms. No lockout.
- **Clock** — `datetime('now')` and `datetime.now(UTC)` are both UTC on the same host.
- **Refusal message** — names the cause and the recovery, and carries no secret.

PR #235, opened 2026-08-05
fix: unknown watts must suppress the net, not price it at zero

GeiserX:
Closes `CashPilot-c6u` (P1), found by the refine sweep and verified at the anchors before filing.

## What was wrong

`/api/earnings/net` answers "what did I actually keep?". Its docstring promises it never presents gross as profit, and `app/power.py` says the same thing twice: *"a zero cost would render gross as net and quietly overstate earnings."*

It did that. `_get_all_worker_containers` reaches the workers over HTTP, so it fails for ordinary reasons — a host down, a proxy blip, a busy SQLite. The handler caught that and substituted `statuses = []`:

```python
except Exception as exc:
    logger.warning("Worker status unavailable for the power estimate: %s", exc)
    statuses = []
```

Downstream, that is indistinguishable from a fleet with nothing running. Every service was charged 0 W, `total_cost` summed to `0.00`, and `total_net` came out **equal to `total_gross`**.

The lie is `cost_known`. `app/power.py:137` derives it from the tariff alone:

```python
configured = price_per_kwh > 0
```

It never asks whether any watts were measured. So with a tariff set, the endpoint reported a cost of zero and a net equal to gross, and labelled it **known**.

This endpoint's own comment already describes that outcome as a bug that bit the project twice, both times through the `service` vs `slug` key. The `except` branch recreates the same precondition by a different route.

## The fix

An empty list from a **failure** is not the same fact as an empty list from an idle fleet. The handler now tracks which one it has, and unknown watts suppress the net through the mechanism **already in this function** for unavailable FX — passing a zero price, which makes `summarise` return `None` for cost and net rather than numbers.

Gross is still reported: it comes from the database and no worker outage can affect it.

| condition | `cost_known` | `total_cost` | `total_net` | `total_gross` |
|---|---|---|---|---|
| worker lookup failed | `false` | `null` | `null` | reported |
| fleet genuinely idle | `true` | number | number | reported |
| no tariff configured | `false` | `null` | `null` | reported |
| FX rate unavailable | `false` | `null` | `null` | reported |

`watts_unavailable` and `fx_unavailable` are separate flags, so the two causes never mask each other, and the reason string rules out the zero reading in words: *"the running cost and the net figure are unknown for this window, not zero."*

## Evidence

11 tests. Four negative controls, all caught:

| reverted | result |
|---|---|
| the original bug — unknown watts priced at the tariff | 2 failed |
| `watts_known` never set false | 7 failed |
| suppressed but unexplained | 2 failed |
| **over-broad** — an idle fleet also reports unknown | 3 failed |

That last one matters as much as the first: it proves the suppression is narrow, and that a genuine measurement of zero draw still produces a net.

Gates: `ruff check` clean, `ruff format --check` clean, both Node harnesses PASS, **3186 passed, 95.34% coverage**.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -2342,6 +2342,11 @@ async def api_earnings_net(request: Request, days: int = 30) -> dict[str, Any]:
     except (TypeError, ValueError):
         host_tdp = power.DEFAULT_HOST_TDP_WATTS
 
+    # Whether the power side of this endpoint has any input at all. An empty
+    # `statuses` from a FAILURE is not the same fact as an empty one from a
+    # fleet with nothing running, and the difference decides whether a net
+    # figure exists (CashPilot-c6u).
+    watts_known = True
     try:
         statuses = await _get_all_worker_containers()
     except Exception as exc:
@@ -2349,6 +2354,7 @@ async def api_earnings_net(request: Request, days: int = 30) -> dict[str, Any]:
         # come from the database and are still perfectly reportable.
         logger.warning("Worker status unavailable for the power estimate: %s", exc)
         statuses = []
+        watts_known = False
     # Only RUNNING containers. A stopped one draws nothing, and counting it
     # inflates a worker's container count, which shrinks every running service's
     # share of that host's idle floor and understates the fleet's real cost.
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #236, opened 2026-08-05
fix: stop asserting a zero nobody measured, and say why Settings is empty

GeiserX:
Closes `CashPilot-pmi` and `CashPilot-cn3` (both P2). Found by the refine sweep, each verified at its anchors before filing.

## pmi — a measured-looking zero

Three of the dashboard's four stat cards initialise to an em-dash. `#active-services` initialised to a literal `0`.

That card is the one that was **specifically fixed for this in JS**:

```js
// `|| 0` would render "could not be counted" as "nothing is running".
setTextContent('active-services', data.active_services == null ? '—' : data.active_services);
```

and the fetch's catch deliberately preserves whatever is on screen:

```js
} catch (err) {
  // Keep whatever was already displayed — a transient fetch failure
  // must not look like the dashboard's earnings dropped to zero.
```

Compose the three and the bug appears only on a **first** load that fails: "whatever was already displayed" is the template's hardcoded `0`. The user is told, in the same typeface as a real measurement, that nothing is running — at the moment nothing was measured.

One character in a template undid a fix that had already been made in JS.

The test that pins it is not "this card says em-dash" — it is **no stat card anywhere may initialise to a numeric literal**, so the next card added is covered too.

## cn3 — panels that spin forever

```js
const [config, envInfo, collectorsMeta] = await Promise.all([
  api('/api/config'),                       // no .catch
  api('/api/env-info').catch(() => []),
  api('/api/collectors/meta').catch(() => []),
]);
...
} catch (err) {
  // Settings may not be available yet
}
```

Two of the three calls defend themselves; `/api/config` does not. Its rejection aborts the whole `Promise.all` before either render function runs, both containers keep the template's `Loading...`, and the empty catch guarantees nothing will ever say otherwise. An expired session, a restarting server, or a network blip is indistinguishable from a page that just never finished loading.

Every other loader in this file writes a reason into its container. This one was the outlier.

Now both panels get a cause and an action, and the server's detail is escaped before it reaches `innerHTML` — it is untrusted input.

## Evidence

`scripts/settings_failure_check.mjs` runs the **real** functions against a stub DOM, and is wired into `test.yml` beside the existing browser-free harnesses. A string assertion can prove the catch is non-empty; only running it proves both panels get written, that the text is useful, and that the detail is escaped.

16 pytest tests plus 9 harness assertions. Six negative controls, all caught:

| reverted | caught by | result |
|---|---|---|
| the hardcoded `0` restored | pytest | 3 failed |
| catch emptied again | pytest | 1 failed |
| only one panel written | harness | FAIL |
| message no longer escaped | harness + pytest | FAIL / 1 failed |
| message reduced to "please try again" | harness | FAIL |
| harness removed from CI | pytest | 1 failed |

One control initially reported NOT CAUGHT and was **my error, not a gap**: the anchor `${escapeHtml(message)}` also occurs in `toast()`, so the patch landed on the wrong function. Re-run with a unique anchor, both the harness and pytest catch it.

Gates: `ruff check` clean, `ruff format --check` clean, `node --check` clean, all three browser-free harnesses PASS, **3191 passed, 95.32% coverage**.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,6 +44,12 @@ jobs:
       - name: Currency formatting behaviour
         run: node scripts/currency_check.mjs
 
+      # Also browser-free. A string test can prove the Settings catch is no
+      # longer empty; only running it proves BOTH panels get written and that
+      # what lands in them says why (CashPilot-cn3).
+      - name: Settings failure path
+        run: node scripts/settings_failure_check.mjs
+
       # Also behaviour pytest cannot reach, and for the same reason: these
       # renderers live in a <script> block inside a Jinja template, so a Python
       # assertion could only prove the source mentions a variable — never what
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -2607,8 +2607,34 @@ const CP = (() => {
       renderCollectors(collectorsMeta, config);
       loadCredentialHealth();
     } catch (err) {
-      // Settings may not be available yet
+      // Say what happened. /api/env-info and /api/collectors/meta each carry
+      // their own .catch, so the only call that can reject is /api/config —
+      // and its rejection aborts the whole Promise.all before either render
+      // runs. Both panels then keep the template's "Loading..." forever, and
+      // this catch used to be empty, so nothing ever corrected it: an expired
+      // session or a restarting server looked like a page that simply never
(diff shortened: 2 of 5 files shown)


GeiserX:
@coderabbitai review

PR #237, opened 2026-08-05
fix: three ways the fleet page misled the person reading it

GeiserX:
Closes `CashPilot-2dh`, `CashPilot-0p4` and `CashPilot-wij` (all P3). All three are in `fleet.html`, which had not had the pass `app.js` received.

## 2dh — a live worker that looks two hours stale

The database stores UTC (`datetime('now')`, SQLite), and the frontend had **no date formatting of any kind**: `toLocaleString`, `toLocaleDateString` and `new Date(` each return **zero** hits across `app.js` and `fleet.html`. So the raw string went straight to the page.

A viewer in CEST reads `Last seen: 2026-08-05 04:00:00` at 06:05 local and concludes the worker has been silent for two hours. It heartbeated five minutes ago. That string is the evidence for *"This host is not reachable"*, and it also appears in the Remove-worker confirmation — acting on a misjudged staleness takes a live host out of the fleet.

One shared `fmtTimestamp` now:

| input | text | title |
|---|---|---|
| `2026-08-05 04:00:00` | local time | `2026-08-05 04:00:00 UTC` |
| `null` | `never` | — |
| unparseable | echoed back verbatim | "could not read this timestamp" |

The `Z` suffix matters: `Date.parse` treats a zone-less string as **local**, so parsing without it would have silently shifted the value in the other direction.

## 0p4 — a Copy button that reported success it never had

```js
navigator.clipboard.writeText(text).then(() => CP.toast('Copied', 'success'));
```

Two failures in one line. `navigator.clipboard` is gated on a **secure context**, and `docs/fleet.md` documents the normal setup as `http://192.168.10.100:8080` — there the object is `undefined`, the property access throws inside an `async` function, and `delegate.js` calls it without awaiting, so the rejection is unhandled. The button was simply dead, silently, for exactly the multi-server users who need it.

And when the owner-only reveal fails, `_apiKey` is the literal `(error)`. The toast said success anyway, so the user pasted `CASHPILOT_API_KEY=(error)` into their compose file and then debugged a worker that could never enrol.

Now: sentinels are refused by name, the write is awaited and caught, and a non-secure origin gets the text selected in a textarea with an explanation — a fallback that cannot itself fail.

## wij — a reboot that reads as lost containers

`api_fleet_summary` skips offline workers **before** accumulating, so of four cards in one row, `Workers`/`Online` are fleet-wide and `Services`/`Running` are online-only. Nothing said so.

The dashboard already ships the right explanation and a **"Check the fleet"** button beneath it:

> Services running on … are not shown here — this does not mean they stopped. Containers keep running and earning while a worker is offline.

The user was routed from a correct explanation to a page that restated the error it was written to prevent. The cards now say `(online)`, and the summary reports `unreachable_containers` / `unreachable_workers` so the page can name what it leaves out — and say explicitly that those containers keep running and earning.

**The counts themselves are unchanged.** A test pins that, so this stays a labelling fix.

## A bonus, found while testing

`fleet_staleness_check.mjs` ended with:

```js
console.log(`fleet staleness check passed (${20} assertions)`);
```

A hardcoded literal. It reported twenty assertions whether twenty ran or two did — a number that looked measured and was not, in the script whose whole job is catching that. It now counts: **28**.

The harness also now extracts the **real** `fmtTimestamp` from `app.js` instead of stubbing it. A stub would let the formatter break while the harness stayed green.

## Evidence

24 pytest tests, 28 harness assertions. Ten negative controls, all caught:

| reverted | caught by |
|---|---|
| raw UTC rendered again | pytest |
| timestamp parsed as local, not UTC | harness |
| a time invented for an unparseable value | harness |
| `(error)` sentinel copied | pytest |
| unawaited write with an unconditional toast | pytest |
| cards unlabelled | pytest |
| endpoint stops reporting what it omits | pytest |
| note never rendered | pytest |
| harness stubs the formatter | harness |
| assertion count hardcoded again | pytest |

All four sources verified byte-identical after every revert. Two of my own tests initially failed by matching their own explanatory text; both were replaced with structural checks.

Gates: `ruff check` clean, `ruff format --check` clean, `node --check` clean, all three browser-free harnesses PASS, **3226 passed, 95.34% coverage**.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -3799,11 +3799,21 @@ async def api_fleet_summary(request: Request) -> dict[str, Any]:
     total_running = 0
     online_workers = 0
 
+    # Containers on workers that are NOT online. They are still running and
+    # still earning -- an unreachable worker is a reporting gap, not a stopped
+    # one -- so the fleet page can say how much of the estate these cards leave
+    # out instead of letting a reboot read as containers lost (CashPilot-wij).
+    unreachable_containers = 0
+    unreachable_workers = 0
+
     for w in workers:
+        _parse_worker_json(w)
         if w["status"] != "online":
+            if w["container_count"]:
+                unreachable_workers += 1
+                unreachable_containers += w["container_count"]
             continue
         online_workers += 1
-        _parse_worker_json(w)
         total_services += w["container_count"]
         total_running += w["running_count"]
 
@@ -3812,6 +3822,8 @@ async def api_fleet_summary(request: Request) -> dict[str, Any]:
         "online_workers": online_workers,
         "total_containers": total_services,
         "running_containers": total_running,
+        "unreachable_containers": unreachable_containers,
+        "unreachable_workers": unreachable_workers,
     }
 
 
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -3238,6 +3238,26 @@ const CP = (() => {
   // -----------------------------------------------------------
   // Public API
   // -----------------------------------------------------------
+  // The frontend had NO date formatting of any kind — grep for toLocaleString,
+  // toLocaleDateString or `new Date(` across app.js and fleet.html returned zero
+  // hits — so every timestamp reached the user as whatever the server
+  // serialised. The DB writes datetime('now'), which SQLite produces in UTC, and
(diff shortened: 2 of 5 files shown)


GeiserX:
@coderabbitai review

PR #238, opened 2026-08-05
fix: the two container inventories no longer disagree

GeiserX:
Closes `CashPilot-dw1` (P3), the last of the seven defects the refine sweep filed.

## What was wrong

`get_status` and `get_status_light` are the same two-phase discovery — label-filtered containers, then image-matched externals — and `get_status_cached` serves **whichever is current for the cache age**.

They disagreed. The light path kept a `seen_slugs` set:

```python
if slug and slug not in seen_slugs:
```

`get_status` had no such set and returned every match. So a host running a CashPilot-managed honeygain alongside a hand-started one reported **two containers or one, depending purely on how warm the cache was** — with nothing in the code saying so.

## Which side is correct

Reporting both:

- the container **IDs differ**, so these are genuinely separate containers, and `seen_ids` already prevents counting one of them twice;
- the UI's per-instance rows exist precisely to show each one;
- two instances of one service on a single host is exactly what a user needs to see, because most providers pay **per IP** — hiding the second hides the reason their earnings are not doubling.

`seen_slugs` is removed rather than left in place unread. A write-only set is how this divergence looked right up close, and leaving it invites the next author to use it again.

## An existing test asserted the old behaviour

`test_image_matched_container_skipped_when_slug_already_seen` pinned the dedupe. It documented the implementation, not a requirement — its own comment is just a restatement of the line above it. It is updated rather than deleted, with the reasoning recorded in the test, and a companion added for the dedupe that **must** stay: the same container surfaced by both scans is still reported once.

## What I deliberately did not change

The sweep also reported that `get_status_light` omits `net_rx_bytes`/`net_tx_bytes` where `get_status` emits them. I checked instead of acting: `net_activity.totals` reads both with `.get` and returns `None` when they are absent, so an omitted key and an explicit `None` are the same answer — *unavailable*. The light path skips `stats()` on purpose because it is slow, so it genuinely cannot know them, and writing `0.0` would be exactly the lie this codebase exists to avoid. Three tests pin that reasoning so the next reader does not have to redo it.

## Evidence

10 new tests plus the updated pair. The central one drives the **same fixture through both functions** and asserts they return the same container IDs — the defect stated as a single assertion.

Two negative controls, both caught:

| reverted | result |
|---|---|
| slug dedupe restored on the light path | 4 failed |
| ID dedupe dropped (one container counted twice) | 2 failed |

Source verified byte-identical after each revert.

Three of my own fixtures were wrong before they were right, each time by guessing instead of reading: `deployed_by` is `worker` not `cashpilot`, and the label is `cashpilot.deployed-by` with a **hyphen**. The fixture now keys off `app.constants` directly so that class of error cannot recur.

Gates: `ruff check` clean, `ruff format --check` clean, all three browser-free harnesses PASS, **3213 passed, 95.42% coverage**.

### DIFF
--- a/app/orchestrator.py
+++ b/app/orchestrator.py
@@ -740,14 +740,12 @@ def get_status_light() -> list[dict[str, Any]]:
         filters={"label": f"{LABEL_MANAGED}=true"},
     )
     seen_ids: set[str] = set()
-    seen_slugs: set[str] = set()
 
     results: list[dict[str, Any]] = []
     for c in labeled:
         try:
             seen_ids.add(c.id)
             slug = c.labels.get(LABEL_SERVICE, "unknown")
-            seen_slugs.add(slug)
             results.append(
                 {
                     "slug": slug,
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #239, opened 2026-08-05
fix: show the version the app is actually running

GeiserX:
Reported from the live UI: the sidebar read **CashPilot v0.2.49** while the running release was **v1.11.34**.

## The cause is one literal

`app/templates/base.html:102`

```html
<span class="sidebar-footer-version">CashPilot v0.2.49</span>
```

Hardcoded, last updated by hand at 0.2.49, roughly **150 releases** ago.

## Everything needed to do it properly already existed, and was correct

| | |
|---|---|
| `Dockerfile:25-26` | `ARG CASHPILOT_VERSION=dev` → `ENV CASHPILOT_VERSION=$CASHPILOT_VERSION` |
| `build.yml` | passes `build-args: CASHPILOT_VERSION=<resolved release>` |
| `app/version.py:26` | `current()` reads that env var |

Verified on the live container before changing anything:

```
env CASHPILOT_VERSION=1.11.34
version.current() = 1.11.34
```

The correct value was sitting in the process the entire time. The template just didn't ask for it.

## Why this isn't cosmetic

1. It is the **only** place a user can see which version they run. Every bug report filed from this UI has carried a version 150 releases wrong.
2. The fleet page already computes version **skew** from `version.current()` — so the sidebar and the fleet page disagreed with each other about what the UI was.
3. It cannot self-correct. A hand-maintained version in a template is guaranteed to drift, and 0.2.49 → 1.11.34 is the proof.

## The change

`version.display()` — a release reads `v1.11.34`, anything else reads what it actually is. Registered as a **Jinja global**, so no handler threads it and a new template can't reintroduce a literal by copy-paste.

It never fabricates a number. Note one deliberate choice: `latest` passes through as `latest` rather than being flattened to `dev`. A container built from the floating tag is not a dev build, and "latest" tells the user something actionable — you're on a moving tag, pin a version. My first test asserted `dev` there and was wrong; the invariant that actually matters is that no non-release ever renders as a version-looking string, and that is what is now asserted.

## The test that matters long-term

Not "this line is right" but **no template anywhere may print a version literal** — parametrized over every `.html` under `app/templates`, with an emptiness guard so the sweep can't go vacuously green. Comment lines are excluded so a template may explain the history without repeating it.

Four negative controls, all caught:

| reverted | result |
|---|---|
| the original hardcoded `v0.2.49` | 3 failed |
| Jinja global unregistered | 2 failed |
| a dev build fabricates `v0.0.0` | 4 failed |
| a release loses its `v` | 2 failed |

Sources verified byte-identical after every revert.

Gates: `ruff check` clean, `ruff format --check` clean, all three browser-free harnesses PASS, **3259 passed, 95.42% coverage**.

## Summary by CodeRabbit

* **New Features**
  * The sidebar now displays the application’s current runtime version automatically.
  * Development and unknown builds display as `dev`, avoiding misleading version numbers.
  * Release versions are formatted consistently with a `v` prefix.

* **Bug Fixes**
  * Removed reliance on a hardcoded version number, ensuring displayed version information stays accurate across builds.

* **Tests**
  * Added coverage for version display in templates, runtime configuration, Docker builds, and the CI workflow.

### DIFF
--- a/app/deps.py
+++ b/app/deps.py
@@ -18,7 +18,7 @@ from fastapi import HTTPException, Request
 from fastapi.responses import RedirectResponse
 from fastapi.templating import Jinja2Templates
 
-from app import auth, setup_token
+from app import auth, setup_token, version
 
 __all__ = [
     "templates",
@@ -51,6 +51,11 @@ def _csp_nonce(request) -> str:
 
 
 templates.env.globals["csp_nonce"] = _csp_nonce
+# Registered as a GLOBAL so every template gets the running version without
+# each handler threading it through, and so a new template cannot quietly
+# reintroduce a hardcoded literal by copy-paste (CashPilot: the sidebar shipped
+# "v0.2.49" for ~150 releases).
+templates.env.globals["app_version"] = version.display
 
 
 def _login_redirect() -> RedirectResponse:
--- a/app/templates/base.html
+++ b/app/templates/base.html
@@ -99,7 +99,7 @@
           <svg width="14" height="14" viewBox="0 0 16 16" fill="currentColor"><path d="M4.25 2.5c-1.336 0-2.75 1.164-2.75 3 0 2.15 1.58 4.144 3.365 5.682A20.6 20.6 0 008 13.393a20.6 20.6 0 003.135-2.211C12.92 9.644 14.5 7.65 14.5 5.5c0-1.836-1.414-3-2.75-3-1.373 0-2.609.986-3.029 2.456a.749.749 0 01-1.442 0C6.859 3.486 5.623 2.5 4.25 2.5z"/></svg>
         </a>
       </div>
-      <span class="sidebar-footer-version">CashPilot v0.2.49</span>
+      <span class="sidebar-footer-version">CashPilot {{ app_version() }}</span>
     </div>
   </aside>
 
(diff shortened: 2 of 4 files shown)


GeiserX:
Both findings were real and are fixed in the latest commit.

**1. `vnot-a-release`** — confirmed by driving it: `is_release()` only excludes a known list, so any arbitrary string got the `v`. Now gated on `series()`, which requires MAJOR.MINOR of digits. Verified across the range:

| `CASHPILOT_VERSION` | renders |
|---|---|
| `1.11.34` | `v1.11.34` |
| `not-a-release` | `not-a-release` |
| `latest` | `latest` |
| unset | `dev` |

**2. Bare literals** — the gap is real, but I did not take the proposed regex. `\d+\.\d+` matches **292** things in these templates (every `1.5rem`, `0.85rem`, opacity value), and even a three-component form matches SVG path coordinates, the `chart.js@4.4.8` CDN pin, and legitimate prose about a minimum worker version. A sweep that cries wolf on 292 lines gets deleted by the next person.

Anchored on the product name instead — `CashPilot\s+v?\d+\.\d+` — which is the actual failure mode. Catches `CashPilot v0.2.49` **and** `CashPilot 0.2.49`, with zero false positives on the five real template lines now pinned as negative cases.

Negative control on (1): restoring `is_release()` fails the suite. 3268 passed, 95.44%.

PR #243, opened 2026-08-05
docs: add UPGRADING.md, and pin the compose examples to 1.12

GeiserX:
Closes `CashPilot-su5` (P1), the half that matters most.

## The problem

The changelog documents **8 of 198 releases**, and its newest entry is `[Unreleased]`. Someone on 1.10.1 — which your fleet was actually running, 33 releases behind — had no way to learn what changed or whether anything needed doing.

Research (recorded on the bead) found the shape every well-run project converges on: a **short, hand-written document listing only the releases that require an action**, separate from the full changelog. Synapse, Zulip and authentik all do this. The changelog is a reference index; `UPGRADING.md` is the one an operator is obliged to read.

## The content was harvested, not invented

Three hand-written `**Upgrade note:**` bullets were buried in the changelog's `[Unreleased]` block. They are the best writing in that file and would have been **lost the moment anything regenerated it** — which is exactly what the automation half of this bead proposes. Rescuing them first was the point.

Every version heading was established by finding the commit that introduced the behaviour and asking git which tag contains it — not guessed:

| release | entry | how the version was established |
|---|---|---|
| **v1.11.30** | enrollment window bounded | first commit touching `WORKER_KEY_CONFIRM_WINDOW` |
| v1.11.4 | compose pins a series | first commit pinning `cashpilot:1.11` |
| v1.5.0 | refuses to start on unwritable `/data` | first commit with `CASHPILOT_ALLOW_EPHEMERAL_KEY` |
| v1.1.0 | `/metrics` can require a token | first commit with `CASHPILOT_METRICS_TOKEN` |
| v1.0.4 | dashboard binds loopback | first commit with `CASHPILOT_BIND_ADDR` |
| v1.0.0 | per-worker fleet keys | links the existing page |

**v1.11.30 had no note anywhere**, and it is the entry most likely to actually disconnect somebody — an unconfirmed worker starts getting 401s 24 hours after enrolment. It explicitly names the Android app before v0.2.0 as an affected case.

## The shape of an entry

Each one answers three questions, in the order a reader needs them: **who is affected** (with an explicit "no action is required" clause so most people can stop reading), **what breaks if you do nothing** in observable terms, then **the command**.

## A real bug this caught

The full suite failed on `test_the_pin_matches_the_newest_released_series`. Releasing **v1.12.0** — the earnings feature bumped the minor — moved the newest series while the compose examples still pinned `1.11`. That test exists because a stale pin gave issue #188 a version with a first-run bug fixed months earlier.

Bumped both files to `1.12`, after verifying **both images exist on Docker Hub** rather than assuming the release published them.

## Tests

28, and three of them failed on the first run against real gaps I'd left:

- **the README didn't link to it** — a guide nobody can find is not a guide
- **the changelog didn't link to it** either
- one "no action" clause was invisible to the check because markdown wrapped the line (my test's bug, not the doc's — now normalises whitespace)

The guards that matter long-term: every version heading must be a **real git tag** (a guide with invented versions is worse than none), every entry must scope who is affected and offer a no-action clause, and the three harvested notes plus the `.fernet_key` backup warning must all still be present.

Gates: `ruff` clean, all four browser-free harnesses PASS, **3336 passed, 95.44% coverage**.

## Not in this PR

The git-cliff automation — generating `CHANGELOG.md` from conventional commits and rendering the same notes into the release body. That is the other half of the bead and is a separate, mechanical change. This one had to land first so the harvested prose is safely out of `[Unreleased]` before anything regenerates it.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated upgrade guide with version-by-version steps and important migration notes.
  * Added upgrade notices in the main documentation to make the guide easier to find.

* **Updates**
  * Updated bundled container images to version 1.12 for the app and worker components.

* **Tests**
  * Added checks to ensure the upgrade guide stays complete, accurate, and linked from the docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 All notable changes to CashPilot are documented here.
 
 **Upgrading?** [UPGRADING.md](UPGRADING.md) lists only the releases that require an
-action, with what breaks if you skip them. This file is the full history.
+action, with what breaks if you skip them. This file carries the detailed notes
+recorded so far; it does not yet cover every release.
 
 ## [Unreleased]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to CashPilot are documented here.
 
+**Upgrading?** [UPGRADING.md](UPGRADING.md) lists only the releases that require an
+action, with what breaks if you skip them. This file is the full history.
+
 ## [Unreleased]
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ The key differentiator: a browser-based setup wizard guides you through account 
 - **Simple two-container setup** -- UI + Worker, no dependencies to install
 - **Service catalog** with earning estimates, requirements, and platform details
 
+> **Upgrading an existing install?** Read [UPGRADING.md](UPGRADING.md) first. It lists only the releases that need you to do something.
+
 ## Quick Start
 
 With Docker Compose (recommended):
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,0 +1,148 @@
+# Upgrading CashPilot
+
+**Read this before upgrading.** Only releases that need you to *do* something are
+listed here. If a version is not mentioned, `docker compose pull && docker compose up -d`
+is all it needs.
+
+The full per-release list of changes is in [CHANGELOG.md](CHANGELOG.md). This file
+is the short one: what breaks, who it affects, and what to type.
+
+**Always upgrade the UI and every worker to the same version.** The compose files
+pin both images to the same tag for exactly this reason.
+
+**Back up `/data` first.** `cashpilot.db` holds your earnings history and
+`.fernet_key` decrypts every stored credential — without that key they cannot be
+recovered. See [Backing up node identities](docs/backup.md) for the rest.
+
+---
+
+## v1.11.30 — the shared fleet key stops working for a worker that never enrolled
+
+**Affects you if** a worker appears in your fleet but has never confirmed its own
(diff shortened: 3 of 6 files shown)


GeiserX:
All nine addressed. Three were serious and one would have caused real damage — thank you.

**The `/data/.worker_id` one is the important catch.** My re-enrolment steps said "remove the worker, it re-enrols" and never mentioned the identity file. Without it the worker registers under a **new** id and you get a duplicate fleet row instead of the same machine returning — the exact 401-lockout trap this project has already hit once. The steps now preserve it, check ownership, and **restart** the container, since the id is read at startup and `up -d` on a running container does not re-read it.

**`CASHPILOT_WORKER_BIND_ADDR`** — correct, and I had conflated it with the UI variable. A fleet's remote worker left on loopback is unreachable and its containers unmanageable. Documented separately, and pointed at a tailnet address rather than `0.0.0.0`, since that port carries the Docker socket API.

**`:latest`** — right, and I should have caught it: the project forbids it precisely because it makes the running version unknowable, which is what produced a months-old bug report against an already-fixed bug. Replaced with an explicit pin.

**The changelog overstatement** — this one was self-contradictory. I wrote "the full per-release list of changes is in CHANGELOG.md" in a PR whose entire premise is that it covers 8 of 198 releases. Fixed in both files.

**The four test findings were all real weaknesses in guards I wrote**, and the tag one is the worst: it *skipped* when no tags were present, so a shallow CI checkout would have passed with invented version headings — defeating the only thing that test existed to prevent. It now fails in CI and skips only outside it.

One change beyond your list: I removed the **v1.1.0** entry entirely. The optional `/metrics` token requires no action, and its presence weakened the promise that everything in this file is something you must do — which is the document's whole value. A test now enforces that promise, and it is what surfaced the problem.

3348 passed, 95.44%.

PR #249, opened 2026-08-05
docs: CLAUDE.md denied a device capability that has shipped since v1.5.1

GeiserX:
Fallout from closing `CashPilot-uld`. Its follow-up asked me to check whether the four GPU services declare a device in their catalog entries — the worry being that they would repeat the Mysterium bug, where a container starts without its device, looks healthy, and **earns nothing**.

They do not declare one. But they cannot hit that bug either, and the check turned up something worse one line away.

## CLAUDE.md was denying a shipped capability

```
The catalog cannot declare devices yet, so a CashPilot-deployed Mysterium
container does not get it automatically.
```

True before the TUN fix, never updated. It is now false three times over:

| Claim | Reality |
|---|---|
| the catalog cannot declare devices | `services/bandwidth/mysterium.yml` → `docker.devices: ['/dev/net/tun']` |
| …so it is not forwarded | `app/main.py:1476` → `"devices": docker_conf.get("devices") or None` |
| …so the container does not get it | `app/worker_api.py:1037-1121` validates against a per-slug allowlist, then passes it to Docker |

`docs/guides/mysterium.md` already says **"Fixed in v1.5.1+."** Only CLAUDE.md disagreed — the one file an agent reads first.

That is why this is worth a PR rather than a one-word edit. **A stale instruction is worse than a missing one, because it gets followed.** Anyone acting on it would either re-implement a capability that already exists, or tell a user to add `--device` by hand for no reason. This project has a rule about not inventing facts; a document asserting a shipped feature is absent is the same failure in prose.

## The security note the bead asked for

Passing a host device widens what a container can touch, so `security-defaults.md` now states the actual rule rather than leaving it in the code: a deploy may request a device only if **its own** catalog entry declares it **and** the device is on a fixed ceiling — today `/dev/net/tun` alone. One service declaring a device does not let another request it, and anything outside the ceiling is refused rather than quietly dropped.

## The tests pin prose to code, in both directions

The interesting part. Rather than checking each file for a string it can satisfy by itself, the tests read `_ALLOWED_DEVICES` **out of the code** and require the docs to agree:

- widen the ceiling in code, say nothing in the docs → **fails**
- advertise a device in the docs that deploys would refuse → **fails**
- the stale CLAUDE.md sentence returns → **fails**
- a compute service gains an image without declaring a device → **fails**

All four verified by mutation. The fourth is why the GPU follow-up closes as a **false premise** rather than a fix: every compute service ships an empty `docker.image`, so none is deployable, and building device plumbing for them now would be speculation. The assertion is conditional on deployability, so the day one gains an image without its device, CI says so.

Worth recording that control 4 initially did **not** fire — I patched `image: ''` while the file says `image: ""`, so the mutation never applied and the control passed vacuously. A control that passes is a control that proved nothing. It asserts the anchor is unique before replacing now.

## Evidence

`ruff` clean, `mkdocs build --strict` clean, **3428 passed, 95.44% coverage**.

## Summary by CodeRabbit

* **Documentation**
  * Updated Mysterium deployment guidance to cover automatic networking-device setup and when older containers must be recreated.
  * Documented service-level device allowlisting and deployment restrictions.

* **Tests**
  * Added coverage confirming device documentation matches enforcement.
  * Verified Mysterium device forwarding and prevented GPU deployments without declared devices.


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #250, opened 2026-08-05
feat(ui): show host disk and GPU in the services table

GeiserX:
Closes the UI half of `CashPilot-cle`. Filed at your request: *"columns in the main table in the web should include disk (for example, storj and others) and gpu (if using) — even if its remote like in cashpilot-desktop"*.

## Why these two

The table showed CPU and memory — the two resources that matter **least** for this product — and neither of the two that decide whether a service earns at all.

| | Why it decides viability |
|---|---|
| **Disk** | Storj is paid for what it **stores**, so free space *is* earning capacity. A node that quietly fills up stops growing, and nothing in CashPilot said so anywhere. |
| **GPU** | Salad, Nosana, io.net and Vast.ai only earn with one. There was no way to tell a GPU service that is **earning** from one that is running and idle because the device was never passed through — healthy-looking, earning nothing. |

Both read from the worker that reported them, so a remote host works exactly like the local one.

## Three decisions where the wrong answer looks fine

**It is HOST free space, not the service's own volume.** Different numbers; showing one under the other's name is worse than showing neither. The header says **"Host disk"** and the tooltip denies the wrong reading outright.

**A multi-host service shows the tightest host, not an average.** An average hides exactly the host about to fill up. Your fleet is the case in point — geiserct at 76% next to watchtower with terabytes free would average to "fine".

**GPU stays three-valued.** `available: null` is *"could not tell"* — what a containerised worker reports when the device was never passed through — and renders as unknown, never `None`. Reporting it as `None` would state as fact that a machine has no GPU while it sits on three idle render nodes, which is how your entire fleet currently reads.

Sorting sinks unknown in **both** directions. A numeric sentinel cannot: `-Infinity` sorts last descending but **first** ascending, which would rank every unreadable host as having the least free space. It is a `Symbol`, handled explicitly in the comparator.

## The bug that justifies the harness

`scripts/host_resources_check.mjs` runs the real functions against worker payloads — 37 assertions, wired into CI. My first draft iterated `svc.instances`:

```js
for (const inst of (svc.instances || svc.details || []))
```

`instances` is a **count**. `for...of` over a number throws, so the entire services table would have gone blank — not one column degrading. It is `instance_details`, verified against `app/main.py:1229` rather than inferred from the name. **No string assertion would have caught this.**

Two of my own test expectations were also wrong and the code was right: I asserted 76% was "near full" when the threshold is 80%, and I looked for `"not visible"` when the reason says `"no GPU is visible"`. Both fixed against the real strings.

## Negative controls

| Mutation | Result |
|---|---|
| report an unchecked GPU as `None` | 3 fail |
| average the disks instead of the tightest | 1 fail |
| unknown free space collapses to `0` | 4 fail |
| drop escaping on the device name | 1 fail |
| treat a failed worker lookup as an empty fleet | **passed — harness was wrong** |

The last one is the useful one. It passed because the harness called `setHosts()` directly and never exercised `buildHostResourceMap`. Fixed, and it now fails as it should. A control that passes proved nothing.

Also covered structurally: header, main row and instance sub-row must agree on column count (all 11) — a mismatch shifts every cell after it and no JS assertion would notice.

## Evidence

`ruff` clean, `node --check` clean, all **five** browser-free harnesses pass, **3451 passed, 95.44% coverage**.

## Not in this PR

Per-service **volume** usage (as opposed to host free space) needs per-volume stats from the worker and is a larger change. The column is labelled so it cannot be mistaken for that in the meantime.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,13 @@ jobs:
       - name: Fleet staleness behaviour
         run: node scripts/fleet_staleness_check.mjs
 
+      # The Disk and GPU columns. Worth running rather than grepping for the
+      # same reason as the rest: the first draft iterated `svc.instances`, which
+      # is a COUNT, and `for...of` over a number throws — a string assertion
+      # would have been perfectly happy while the services table went blank.
+      - name: Host resource columns
+        run: node scripts/host_resources_check.mjs
+
       # From the LOCKFILE, which is what Docker ships (`uv sync --frozen` in the
       # Dockerfile) and what release.yml verifies against. Installing from the
       # unpinned requirements.txt meant the suite ran against a resolution
--- a/app/static/js/app.js
+++ b/app/static/js/app.js
@@ -595,6 +595,19 @@ const CP = (() => {
           va = parseFloat(a.memory) || 0;
           vb = parseFloat(b.memory) || 0;
           break;
+        // Both sort UNKNOWN last in either direction rather than as zero.
+        // `parseFloat(x) || 0` — the idiom used above — would rank a host whose
+        // free space could not be read as the fullest disk on the fleet, which
+        // is the same "absent read as a definite value" this codebase keeps
+        // having to undo.
+        case 'disk':
+          va = sortableFreeBytes(a);
+          vb = sortableFreeBytes(b);
+          break;
+        case 'gpu':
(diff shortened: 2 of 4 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #256, opened 2026-08-06
feat(fleet): accept a paired client's own earnings history

GeiserX:
## What

`POST /api/workers/earnings-import` lets a client that has been reading a provider account by itself hand that history to the server. CashPilot Desktop is the case that prompted it: run standalone for months, pair, and the fleet view started on the day of pairing — every earlier day was gone from the total.

This is the second half of `CashPilot-Desktop-xjr`. The first half (#255) added the `source` column and made both delta readers key on `(platform, source)`; this adds the write path that uses it.

## Why separate series, rather than merging into one history

Earnings are stored as cumulative **balance readings**, and an earned figure is the clamped delta between consecutive readings. Interleaving two samplers of the same provider account is therefore not harmless: sampler A reads 12.0, sampler B reads 11.8 an hour later, and the negative step clamps to zero. Every crossing loses a real gain and the total comes out systematically understated.

Storing each client's readings under its own `source` means each series is differenced on its own, and the results are summed. Nothing merges, nothing is double-counted, and unlinking a client leaves its own series intact — which is what makes "show only what this machine earned alone" possible later.

## Security

Two properties, both tested with negative controls:

- **The source comes from the authenticated worker (`cid`), never from the request body.** The model carries no `source` field at all; a test sends one anyway, so a future model change that started honouring it fails here rather than silently letting a client overwrite the server's own series.
- **Only a fully enrolled worker may import** (`state == "ok"`). `enroll` and `reissue` both mean the caller presented the *shared* key, which every worker holds — accepting it would let anyone with that token write a history for any `client_id` they named. The 403 says how to proceed rather than just refusing.

Unknown slugs are skipped and **reported** rather than stored: storing one would create a platform the catalog cannot name, render, or ever collect for again, and a silent drop looks identical to a successful import.

Idempotent by construction — the `(platform, source, date)` unique index means a re-pair or a retried import updates a day rather than adding a second reading for it, which would difference against itself and read as zero.

## Verification

- 14 new tests, all six properties confirmed by **negative control**: hardcoding `source="server"`, allowing shared-key holders, storing unknown slugs, dropping the currency normalisation, coercing an absent FX rate to `0.0`, and accepting a blank `client_id` each fail the tests that claim to catch them.
- Full suite: **3577 passed**, 6 skipped, coverage **95.51%** (gate 90%).
- `ruff check` + `ruff format --check` clean; `node --check`, `currency_check.mjs`, `fleet_staleness_check.mjs` all pass.

Docs: `docs/fleet.md` gains the endpoint, the payload shape, and the three deliberate constraints.

## Summary by CodeRabbit

- **New Features**
  - Added an authenticated earnings-history import endpoint for workers.
  - Supports validated dates, currencies, exchange rates, bounded batch sizes, and duplicate-safe daily updates.
  - Filters unknown catalog entries and reports skipped items.
  - Preserves source isolation and associates imports with the authenticated worker.

- **Bug Fixes**
  - Improved request-validation error serialization, including safe handling of non-finite numeric values.
  - Added atomic batch processing to prevent partial imports after failures.

- **Documentation**
  - Added usage guidance and validation details for earnings-history imports.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -8,10 +8,12 @@ fallback to ./data/cashpilot.db for development.
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json
 import logging
 import math
 import os
+from collections.abc import Mapping, Sequence
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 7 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
Self-review of this diff found two real defects, both now fixed in `f938d6a`:

**A `NaN` balance was accepted and stored.** JSON has no `NaN` or `Infinity`; Python's parser accepts them anyway. One such reading poisons every delta taken from that series — `NaN - x` is `NaN`, and every comparison against it is False, so the clamp misbehaves silently — the account total becomes `NaN`, and serialising that back out emits a bare `NaN` that `JSON.parse` rejects. **One bad reading from one client breaks the dashboard for everyone.** Confirmed by driving the real request before fixing it: `STORED balance: nan  fx: inf`.

**Rejecting one was then a 500.** FastAPI's 422 body echoes the offending input, so the rejection itself could not be serialised. A `RequestValidationError` handler now renders non-finite floats as their names — diagnostic rather than dropping the field — which fixes the class, not just the one endpoint that takes a float today.

My first version of that handler broke every *other* validation error (a custom validator's error carries the raised `ValueError` **object** in `ctx`, so skipping `jsonable_encoder` turned each into a 500). The date tests caught it; it now has its own regression test.

Also deduplicated `skipped`: 400 days of one unrecognised platform returned the same name 400 times.

One negative control **passed**, which meant the code was wrong to have the guard rather than the test being wrong: the sanitiser's `bool` branch was dead (`bool` subclasses `int`, not `float`), so it is gone along with the comment that justified it incorrectly.

3607 tests pass, coverage 95.53%. Five new negative controls, each failing the test that claims to catch it.

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
All three addressed in `7d6bc40`.

**Batching** — implemented as `upsert_earnings_many` (one `executemany`, one commit). One correction to the diagnosis, because it matters for whoever reads this next: `_get_db` hands out a borrowed handle on a **shared** per-loop connection whose `close()` is a documented no-op (`app/database.py` `_BorrowedConnection`), so the loop was never opening and closing a thousand connections. It was **committing** a thousand times — and that half is entirely real, since every commit is an fsync holding SQLite's write lock against this server's own collector.

Batching then introduced a bug of its own, which the new tests caught: the shared connection outlives the request, so a failed batch left an abandoned transaction holding the write lock and **the next write blocked for twelve seconds** before timing out. It now rolls back.

**Docs** — `docs/fleet.md` gains the `Authorization` header and a table of every status the endpoint answers with (200/400/401/403/422).

**Tuple assertions** — fixed at both sites, plus a third the diff didn't touch (`test_optional_runtime.py:38`).

One correction there too, and it changes what the right fix is. The finding says Ruff's `B018` catches this when bugbear is enabled. **Bugbear is already enabled in this repo** (`B` is in `select`; `B018` is not ignored) and ruff 0.15.14 passes the pattern clean — checked against a minimal probe rather than assumed:

```python
m = Mock()
m.assert_not_called(), "this message is dead"
```
→ `All checks passed!`, exit 0.

So nothing in CI would have caught a recurrence. There is now an AST-based test that does — structural rather than a string search, because a string search would match the pattern inside its own docstring.

**One negative control passed, which meant my test was wrong, not the code.** The transactionality test raised while *building* the rows, before any SQL ran — so it proved only that the row build validates first, and it passed against a writer that committed after every row. It now fails inside the statement. Both it and the wedged-connection test fail under their controls.

3619 tests pass, coverage 95.55%.

GeiserX:
@coderabbitai review

PR #257, opened 2026-08-06
docs: group the top-level nav and build the site before merging, not after

GeiserX:
The second half of `CashPilot-gan`. The first half fixed the orphaned guide and generated the guide nav from the catalog; this does the two things left.

## Eleven tabs

`navigation.tabs` is enabled, so **every top-level nav entry is a tab across the header** — there were eleven, with "Network Isolation" sitting as a peer of "Home" and no order a reader could reason about.

Now six: **Home · Getting Started · Service Guides · Configuration · Security · Operations · Reference**.

### The research the bead asked for, before restructuring

The bead sketched "Start here / Run it / Operate it / Reference". That is not what comparable self-hosted projects ship:

| Project | Top-level shape |
|---|---|
| **Paperless-ngx** (Material/Zensical, closest in size) | 11 entries, near-flat, **one** nested group; ordered index → setup → usage → configuration → Administration → advanced → api → development → faq → troubleshooting → changelog |
| **Immich** (Docusaurus, autogenerated) | overview, install, features, administration, guides, developer, api, FAQ |
| **Authentik** (Docusaurus) | core, install-config, add-secure-apps, customize, sys-mgmt, security, troubleshooting, releases, enterprise, developer-docs |

Three things all three agree on, and none of them is the sketched shape:

1. **Order follows the reader's journey** — learn, install, use, configure, operate, reference — not the alphabet.
2. **Section names are domain nouns** (Install, Administration, Security, Reference), not imperative task labels.
3. **Nothing is nested alone.** A group exists because it has siblings.

So: Service Guides sits **third**, straight after Getting Started — it is 50 of the site's ~62 pages and the reason most visitors arrive — and Reference is last. Every group holds at least two pages, which a test enforces.

The two `research/` pages were orphans reachable only by URL. They are the measurements behind catalog thresholds, not drafts, so they are now under Reference.

## The build ran after the merge, not before

`docs.yml` triggered **only on push to main**, and its only step was `mkdocs gh-deploy --force`. So the deploy was the first thing to evaluate the nav: an entry pointing at a file that did not exist would have been discovered by publishing it.

- `mkdocs build --strict` now runs on pull requests touching `docs/`, `mkdocs.yml`, or the workflow.
- `deploy` `needs: build`, so main cannot publish a site the build is failing on.
- Workflow-wide `permissions` dropped to `contents: read`; only `deploy` holds `write`, and it is guarded by `if: github.event_name == 'push'` — a fork PR can run the build but can never publish.

## Verification

- `mkdocs build --strict` passes locally against this nav (real build, mkdocs 1.6.1 + the repo's own `docs/requirements-docs.txt`).
- `scripts/sync_docs_nav.py --check` still round-trips — the generator's splice survives Service Guides moving position.
- 14 new tests + the existing 18 pass.
- **Eight negative controls**, each failing the test that claims to catch it: the old flat nav, a single-page section, a nav entry pointing at a missing file (which `--strict` also refuses), `--strict` dropped, `needs: build` dropped, the push-only guard dropped, write permission granted workflow-wide, and the `pull_request` trigger removed.

One orphan is deliberately left: `docs/growth-strategy.md`. It is an internal promotion playbook ("Track status of each action here"), so adding it to the nav is probably the wrong fix — filed as its own bead rather than decided here.

## Summary by CodeRabbit

* **Documentation**
  * Reorganized documentation into Configuration, Security, Operations, and Reference sections for easier navigation.
  * Added research pages covering idle network traffic and per-IP device limits.
  * Improved documentation validation and publishing workflows to help ensure updates build successfully before release.
* **Quality Improvements**
  * Added automated checks to verify navigation structure, page availability, and safe documentation publishing.


GeiserX:
Fixed in `205338b`. The finding was right, and I could demonstrate rather than argue it: against a `docs.yml` with the deploy condition inverted and an extra write scope added, the old assertions returned

```
'push' in condition          -> True   (condition: github.event_name != 'push')
permissions.contents == read -> True   (permissions: {contents: read, packages: write})
```

— so a workflow that deploys on **every pull request** and grants an extra write scope passed both.

- The strict-build check scanned *every* job's steps, so moving the strict build into `deploy` (which runs only on main — i.e. putting the gate back *after* the merge) kept passing. Now scoped to `jobs.build.steps`.
- The deploy guard used a substring. Now the exact condition.
- The permissions check looked only at `contents`. Now the exact maps, both of them.

Four controls, one per inversion — strict build moved to deploy, condition inverted, an extra write scope, and deploy needing a different job. All four now fail; **three of them passed before this change.**

Also cherry-picked the compose pin bump (`522cff1`) that was failing `test` on this branch. It is pre-existing drift unrelated to these docs changes — v1.15.0 released and the example compose files still pinned 1.15's predecessor — and it fails CI on every open branch until it lands. Filed as `CashPilot-9fg` so the release automation bumps it in future rather than tripping whoever opens the next PR.

PR #260, opened 2026-08-06
fix(release): bump the compose pins through a PR, because main is protected

GeiserX:
The v1.16.0 release **proved the mechanism cannot work**, which is the useful outcome.

`main` is protected — *"Changes must be made through a pull request"* — so the bump was rejected three times with `GH006` and the step warned:

```
! [remote rejected] HEAD -> main (protected branch hook declined)
push rejected (attempt 3); rebasing onto the current branch
::warning::could not push the compose pin bump to 1.16
```

The retry-then-warn design is exactly what let **v1.16.0 ship green** instead of going red at the last step, and that half was right. But branch protection isn't going away, so the *mechanism* had to change rather than the retry count.

## What it does now

Creates a branch, opens a PR, merges it. That works here without weakening anything — `main` requires a PR but demands **no approvals and no status checks**, verified against the protection API rather than assumed. If that ever tightens, the merge fails, the warning fires, and the release still ships. That's the behaviour worth keeping.

Each of the three delivery steps — push, create, merge — warns and `exit 0`s on failure. The validation *above* them still uses `::error::`, because a `sed` that didn't do what it claims is a defect rather than an environment refusal.

Also bumps the pins to **1.16** by hand, which is what the failed run should have done, so `test_compose_image_pins.py` stops going red on every open branch.

## Verification

Five negative controls, all failing:

| Control | Test that fails |
|---|---|
| Back to a direct push to main | `test_it_goes_through_a_pull_request`, `test_it_never_pushes_straight_to_the_release_branch` |
| `pull-requests: write` removed | `test_the_job_may_open_a_pull_request` |
| `GH_TOKEN` removed | `test_the_step_has_a_token_to_talk_to_the_api` |
| A delivery failure escalated to `::error::` | `test_every_delivery_failure_warns_rather_than_failing_the_release` |
| PR base hardcoded to `main` | `test_the_pr_targets_the_branch_the_release_ran_on` |

3653 tests pass, coverage 95.55%.

## Summary by CodeRabbit

* **New Features**
  * Release updates now propose compose configuration changes through a pull request instead of directly modifying the release branch.
  * Release publication can complete successfully even when automated configuration delivery requires manual follow-up.

* **Updates**
  * Updated the CashPilot UI and worker container images to version 1.16 across standard and fleet deployments.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -221,6 +221,9 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      # For the compose-pin bump, which must go through a PR because main is
+      # protected. Nothing else in this job needs it.
+      pull-requests: write
     steps:
       - uses: actions/checkout@v7
         with:
@@ -257,6 +260,10 @@ jobs:
       # a guard on the release. Now the release moves the pin itself, straight
       # after the tag it is pinning to. (CashPilot-9fg.)
       - name: Bump the example compose pins to this series
+        env:
+          # gh needs a token; the job already holds contents: write, and
+          # pull-requests: write is added on the job below for the PR itself.
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
           TAG="${{ needs.release.outputs.new_tag }}"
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #261, opened 2026-08-06
feat(db): say what the schema is and what just migrated

GeiserX:
Closes `CashPilot-sfbh`.

## The gap

Every schema change in `init_db` is guarded by its own `PRAGMA table_info` check, and they're careful. But **nothing said what happened.** An operator watching `docker logs` through an upgrade couldn't tell a clean start from one that had just rewritten the earnings table — and neither could a bug report.

Startup now says:

```
Schema now at version 10 (was 0). Migrations applied this boot: earnings.fx_rate_usd, earnings.source
Schema at version 10; no migration needed this boot.
```

## The design decision that matters

**The version is reported, never used as a gate.**

If `user_version` decided whether migrations ran, a database whose version said `10` but was *missing a column* — an interrupted upgrade, a restored backup, a hand-edited file — could never be repaired, because the gate would say there was nothing to do. Those are precisely the situations where repair matters most.

The `table_info` guards stay the source of truth; they're idempotent and cheap. A test enforces this by handing `init_db` a database that **claims the current version while missing two columns** and asserting both get added.

## It immediately found something

That's the point of the feature, and it paid for itself before merge. A brand-new database reported:

```
Migrations applied this boot: users.password_changed_at
```

`_SCHEMA` never declared that column, so **every fresh install ran an `ALTER TABLE` it never needed**. Declaring it in the `CREATE TABLE` fixes it, and is safe both ways: `CREATE TABLE IF NOT EXISTS` is a no-op on an existing database, which still gets the column from the migration.

Worth distinguishing from a mistake made twice earlier in this work: adding an **INDEX** to `_SCHEMA` is dangerous, because indexes *do* run against existing tables and fail on a column they don't have yet. A column in a `CREATE TABLE IF NOT EXISTS` has no such hazard.

## Verification

Four negative controls, all failing:

| Control | Test that fails |
|---|---|
| Gate the migrations on the version | `test_a_column_is_added_even_when_the_version_already_says_current` |
| Remove the log line | `test_it_always_says_something` + 2 others |
| Never write `user_version` | `test_a_fresh_database_ends_at_the_current_version` |
| Never populate `applied` | `test_an_upgraded_database_names_the_migrations_that_ran` |

3661 tests pass, coverage 95.51%.

## Summary by CodeRabbit

* **New Features**
  * Added database schema version tracking and reporting.
  * Startup now records applied migrations and clearly indicates when no changes were needed.
  * Added support for tracking when user passwords were last changed.

* **Bug Fixes**
  * Database initialization now repairs missing migrations even when the recorded schema version is current.

* **Tests**
  * Added coverage for version persistence, migration logging, repeat initialization, and migration recovery.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -387,7 +387,18 @@ CREATE TABLE IF NOT EXISTS users (
     username   TEXT    NOT NULL UNIQUE,
     password   TEXT    NOT NULL,
     role       TEXT    NOT NULL DEFAULT 'viewer',
-    created_at TEXT    NOT NULL DEFAULT (datetime('now'))
+    created_at TEXT    NOT NULL DEFAULT (datetime('now')),
+    -- Declared here as well as in the migration below, so a FRESH install does
+    -- not run an ALTER TABLE it never needed. Found by the startup log added in
+    -- CashPilot-sfbh: a brand-new database reported "Migrations applied this
+    -- boot: users.password_changed_at", which is exactly the noise that makes an
+    -- operator stop reading migration output.
+    --
+    -- Safe both ways: CREATE TABLE IF NOT EXISTS is a no-op on an existing
+    -- database, so it still gets the column from the migration. (Unlike adding
+    -- an INDEX here, which runs on existing tables and fails on a column they do
+    -- not have yet -- that mistake broke the upgrade path twice.)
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #263, opened 2026-08-06
feat: tell the operator when a newer release exists

GeiserX:
Closes `CashPilot-w0ss`. Your own fleet is the evidence: **33 releases behind, and no surface said so.**

## What was actually missing

Half the machinery already existed — `version.skewed()` compares this build against each worker. What was missing is a reference point *outside* the deployment: nothing in the application had any idea what the newest published release was.

Once a day, one unauthenticated `GET` to the public releases endpoint. A dismissible banner when there's something newer.

## Three constraints, and they're the whole design

- **Silent when unknown.** Offline, firewalled, disabled, never-run, or a dev build all produce `known: false`, and unknown renders **nothing** — no error, no spinner, and above all no reassuring *"up to date"* it never earned. Every failure mode inside `refresh()` lands on the same answer, which is what makes it safe on an air-gapped install.
- **Never auto-updates.** This deploys containers and holds credentials. It tells you; you decide.
- **Answerable.** `CASHPILOT_UPDATE_CHECK=off` makes **no connection at all** — not "fetch and discard", which is precisely what an operator disabling it would assume it *doesn't* do. Tested by asserting the HTTP client is never even constructed.

Dismissible **per version**: dismiss `v1.20.1` and it stays gone until `v1.20.2` exists, so it can't become wallpaper.

## Two existing tests failed, and both were right to

**The scheduler test** pins the exact job set and a uniform `misfire_grace_time`. I'd used `3600` for a daily job. I conformed to `300` rather than carve out an exception to a convention a test enforces uniformly — a missed daily check simply happens tomorrow.

**`test_no_telemetry_or_phone_home_code_exists`** fired on the word *"telemetry"* in my module's docstring — **in a sentence saying it sends none.**

That's worth more than a reword. A check that can't tell a denial from an admission gets worked around by rephrasing, and then it protects nothing. It now strips comments and docstrings but **keeps string literals**, so a telemetry endpoint URL or a `posthog` import is still caught.

Proved by control rather than asserted — dropping this into `app/`:

```python
import posthog
TELEMETRY_URL = "https://example.invalid/collect"
```

fails it on **both** needles.

## What it sends

Nothing about you. No request body, no identifier, and your version is not reported upstream. GitHub learns that an IP asked what the latest release is. Documented in `docs/configuration.md` alongside the opt-out.

## Verification

Five controls on the feature, all failing as required: reporting up-to-date when unreachable; comparing series as **strings** (`1.9` vs `1.10` — live, this project passed its tenth minor); fetching despite the off switch; trusting a non-version tag; re-fetching after every failure.

3688 tests pass, coverage 95.54%.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -52,6 +52,7 @@ from app import (
     preflight,
     producer_state,
     setup_token,
+    update_check,
     version,
 )
 from app.worker_proxy import _pin_url_to_ip, _validate_worker_url
@@ -888,6 +889,21 @@ async def lifespan(app: FastAPI):
         id="db_vacuum",
         max_instances=1,
         coalesce=True,
+        misfire_grace_time=300,
+    )
+    # Once a day. A release is not urgent, and a self-hosted app polling a third
+    # party more often than that is rude. Every failure mode inside refresh()
+    # lands on UNKNOWN, so a firewalled install produces no noise at all.
+    scheduler.add_job(
+        update_check.refresh,
+        "interval",
+        hours=24,
+        id="update_check",
+        max_instances=1,
+        coalesce=True,
+        # 300 like every other job here rather than a carve-out. A missed daily
+        # check simply happens tomorrow; that is not worth an exception to a
+        # convention a test enforces uniformly.
         misfire_grace_time=300,
     )
     scheduler.add_job(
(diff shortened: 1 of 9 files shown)


GeiserX:
@coderabbitai review

PR #265, opened 2026-08-06
docs: say why the UI and worker have separate data volumes, and enforce it

GeiserX:
Closes `CashPilot-4gxr`.

## The boundary

The shipped compose files already gave the UI and the worker separate `/data` volumes. **Nothing said why, and nothing stopped anyone merging them.**

The worker mounts the Docker socket — root on the host. The UI's `/data` holds `cashpilot.db` and `.fernet_key`: the credential store and the only key that can decrypt it. The component that can start any container on the machine has no business being able to read every provider password you've entered.

Consolidating the two is the obvious simplification when tidying a compose file, it looks harmless, and it silently removes that. It now fails a test.

## The mirror is tested too

`/fleet` must stay **shared** — it holds the enrolment key both sides need by definition. A rule applied without its reason gets over-applied, and "separate everything" would break enrolment.

And so is the **premise**: if the UI ever gains the Docker socket, the whole argument stops holding. That is now where it gets noticed.

## One test passed before the work it described existed

`test_the_docs_explain_it` asserted that `"separate"` and `"/data"` both appeared somewhere in `fleet.md`. It went green **before a word of the new section was written** — it was matching *"Comma-separated"* in an environment-variable table.

That's the same near-miss as a negative control that passes, and it's worth naming: a test that goes green too early is telling you something, exactly like one that stays green under a mutation. It now asserts the heading, the volume name, and the phrase that carries the reason.

## Controls

| Control | Test that fails |
|---|---|
| Collapse the two volumes into one | `test_the_ui_and_worker_use_different_data_volumes` |
| Un-share `/fleet` | `test_both_mount_the_same_fleet_volume` |
| Give the UI the Docker socket | `test_the_worker_is_the_one_with_the_docker_socket` |
| Remove the explanation from the compose file | `test_the_compose_file_explains_it` |
| Remove the docs section | `test_the_docs_explain_it` |

3690 tests pass, coverage 95.51%.

### DIFF
--- a/docker-compose.fleet.yml
+++ b/docker-compose.fleet.yml
@@ -85,6 +85,18 @@ services:
       # too. Pick ONE; do not set both.
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
+      # SEPARATE FROM THE UI'S /data, DELIBERATELY. This container has the
+      # Docker socket, which is root on the host. The UI's /data holds
+      # cashpilot.db and .fernet_key -- the credential store and the only key
+      # that can decrypt it -- and the component that can start any container
+      # on the machine has no business being able to read them.
+      #
+      # /fleet IS shared on purpose: it holds only the shared enrolment key,
+      # which both sides need by definition.
+      #
+      # Consolidating these into one volume is the obvious "simplification" and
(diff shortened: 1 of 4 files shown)


GeiserX:
@coderabbitai review

PR #267, opened 2026-08-06
feat(catalog): declare where every service actually pays out

GeiserX:
The catalog foundation for `CashPilot-luj` (tier 1 of the wallet work). Not the screen yet — this is the data model the screen needs, and it's the half that has to be right.

## The problem

A user could not answer the most basic financial question about their own installation: **which addresses do my services pay to?** That lived only inside container env vars and config files, scattered across every worker.

## The design decision

**The model is a first-class field, not an address that happens to be empty.** Three real mechanisms exist, and they aren't variations of one thing:

| Model | Meaning |
|---|---|
| `external` | The user supplies an address they own. There **is** one to show. |
| `internal` | The service holds an off-chain balance and pays on request. There is **no** address until withdrawal — a blank field here reads as *"you forgot to configure this"* when nothing is configurable. |
| `minted` | The container generates its own wallet and accrues to it. The user has typically never seen it. (Tier 3.) |
| `unknown` | Not yet determined — **deliberate, not a placeholder.** The acceptance criteria ask for exactly this: *"visibly flagged, not silently blank."* |

Today: **5 external, 11 internal, 3 minted, 31 unknown.** The unknowns are honest, and they're what the screen will flag.

## Three cases a careless bulk edit would flatten

Each pinned by its own test:

- **Mysterium** is `external`, but its beneficiary is set *after deploy* via TequilAPI — there is no env var to read, and claiming one would show a permanent blank.
- **Traffmonetizer** declares **no chain**. It's USDT, but the chain is chosen per withdrawal, so there is no persistent on-chain identity to name. Naming one would be a fabrication.
- **Salad** is `chain: none` — PayPal and gift cards only — so tier 2 must never try to reconcile it on-chain.

And the one line in here worth real money: **URnetwork forfeits payout after 30 days** if no wallet is linked.

## No private keys

Enforced by a test rejecting `private_key`, `mnemonic`, `seed_phrase`, `keystore` and friends anywhere in a payout block. That custody boundary is *why this tier comes first*: most of the "where is my money" answer at zero custody risk.

## One of my tests failed and was right to

`test_an_address_env_actually_exists_on_the_service` looked for declared env vars under `credentials.fields` and `docker.environment` — **neither exists in this schema.** It found nothing and failed `storj`, the one service that genuinely declares `WALLET`. The lookup was wrong, not the assertion; it now reads `docker.env`.

Worth noting because a laxer version would have passed silently and let a future service name an env var the container never receives.

## Verification

357 new assertions; 4074 tests pass overall, coverage 95.54%.

Five controls, all failing: a missing payout block, an `address_env` naming a var the container never gets, an `internal` service claiming an address, Salad's `chain: none` removed, and a private key added.

## Next

The screen itself, and reading actual addresses from deployed specs.

### DIFF
--- a/services/_schema.yml
+++ b/services/_schema.yml
@@ -136,6 +136,50 @@
 #   currency: USD
 #   notes: "Additional cashout instructions or caveats"
 
+# payout:
+#   WHERE THE MONEY ACTUALLY GOES. `cashout` says how to WITHDRAW; this says
+#   where the funds live and whether there is an address at all (CashPilot-luj).
+#
+#   model: external | internal | minted | unknown        (REQUIRED)
+#     - external: the USER supplies an address they already own. CashPilot never
+#         holds a key. There IS an address to show.
+#     - internal: the service keeps an off-chain balance and pays out on
+#         request. There is NO address to register, and no chain to watch, until
+#         the user withdraws. Showing a blank address field here would read as
+#         "you forgot to configure this" when nothing is configurable — which is
+#         exactly why the model is a first-class field rather than an empty slot.
+#     - minted:   the container generates its own wallet on first run and
+#         accrues to it. The user has typically never seen the address, let
+#         alone backed up the seed.
+#     - unknown:  not yet determined. A DELIBERATE value, not a placeholder: the
+#         registry must flag what it does not know rather than render a blank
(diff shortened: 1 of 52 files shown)


GeiserX:
@coderabbitai review

PR #270, opened 2026-08-06
feat: the payout screen — where your money actually goes

GeiserX:
Completes **CashPilot-luj tier 1**. #269 landed the API; nothing rendered it, so a user still could not answer *"which addresses do my services pay to?"* without reading env vars on every worker.

> Stacked on #269. GitHub will retarget this to `main` when that merges.

## The rule this page holds

Four payout states are genuinely different and must never render alike:

| State | Renders as | Why |
|---|---|---|
| `external` + an address | the address, monospace | it is known |
| `external` + nothing | **"Not set"**, warning-coloured | ACTIONABLE — money may be going nowhere |
| `internal` | *"Held by the service until you withdraw"*, muted | there is **no** address by design; a blank would read as "you forgot something" |
| `minted` | *"The container generates its own wallet"*, muted | the user has probably never seen it |
| `unknown` | *"Not classified yet"*, muted | **our** gap in the catalog, never phrased as user error |

Only a **deployed** service missing an address is highlighted. An undeployed one is information, not a problem to solve today — otherwise choosing what to run puts a permanent nag on your dashboard.

Owner-only, matching `/api/payout-registry` exactly. A laxer gate would render a screen that 403s against its own data, and would advertise to a viewer that the owner's wallet map exists.

## Verification

`scripts/payout_registry_check.mjs` runs the **real renderers** rather than grepping the template — a string assertion could only prove a literal exists, never which state produces it. Wired into `test.yml`.

Mutation-tested in both directions, because a harness that passes first time deserves suspicion:

- making `internal` render as `"Not set"` → **4 checks fail, exit 1**
- removing escaping → **3 checks fail**
- restored → passes, exit 0

Addresses come out of a deployed container's environment, so everything interpolated is escaped; the harness asserts a hostile name and address survive as entities.

Full suite: **4112 passed**, coverage **95.56%**, `app/routers/pages.py` at 100%.

## Found while building this, filed not fixed

`CashPilot-yhkl` — the sidebar has **never** highlighted the current page. `.sidebar-link.active` is styled and every link compares `active_page`, but nothing ever sets it. Left out of this PR because the one-line fix changes five other pages and belongs in its own diff.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,13 @@ jobs:
       - name: Host resource columns
         run: node scripts/host_resources_check.mjs
 
+      # The four payout states. The regression this guards is "internal" (there
+      # is no address, by design) rendering like "not set" (money may be going
+      # nowhere). Both are a cell with no address in it; only the output tells
+      # them apart, so only running the renderer can check it.
+      - name: Payout destination states
+        run: node scripts/payout_registry_check.mjs
+
       # From the LOCKFILE, which is what Docker ships (`uv sync --frozen` in the
       # Dockerfile) and what release.yml verifies against. Installing from the
       # unpinned requirements.txt meant the suite ran against a resolution
--- a/app/routers/pages.py
+++ b/app/routers/pages.py
@@ -54,6 +54,23 @@ async def page_settings(request: Request):
     return deps.templates.TemplateResponse(request, "settings.html", {"user": user})
 
 
+@router.get("/payouts", response_class=HTMLResponse)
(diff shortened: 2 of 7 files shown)


GeiserX:
Superseded by a rebuilt branch — see below.

This PR was stacked on #269, and merging #269 with `--delete-branch` removed its base branch, which GitHub auto-closes a stacked PR for. Nothing was wrong with the change itself.

Rebuilt cleanly on top of the new `main` (the page commit cherry-picked onto the merged API) rather than force-pushing this branch.

PR #273, opened 2026-08-06
feat: read on-chain balances, honestly (dv6 tier 2, first slice)

GeiserX:
First slice of **CashPilot-dv6**. Handles no private key, builds no transaction, signs nothing — the only operation is *"what is the balance at this public address?"*.

## The rule it enforces

The one this codebase already lives by: **absent is not zero.**

A chain we could not reach must never render as a balance of `0`. Zero is a fact about an address; unreachable is a fact about *us*. Showing the second as the first tells someone their money is gone when a public RPC merely rate-limited us.

So every read returns a **state**, and `amount` is `None` unless the state is `known`:

| State | Meaning |
|---|---|
| `known` | we asked and got an answer — **`0` included, because a real zero is a fact** |
| `unreachable` | the RPC failed, timed out, or answered nonsense |
| `unsupported` | no keyless endpoint is configured for that chain |
| `invalid` | the address is malformed, so we never asked |

A caller that formats `amount` without checking `state` gets an em dash, not a figure. That is deliberate.

## Scope and constraints

- **Ethereum, Polygon, Solana** — the only chains any catalogued service declares. Adding an endpoint we have no use for is a dependency on someone else's free infrastructure for no benefit.
- **Keyless by choice.** No shared key ships, and a test asserts every configured endpoint is keyless and HTTPS.
- **`httpx` only**, no chain SDK — the image budget is untouched.
- **`Decimal`, not float.** 18 decimals of wei does not survive a float, and a balance wrong in the last places is a balance nobody trusts.
- **Cached, and backed off per-endpoint.** Public RPCs are a courtesy; a dashboard that polls hard gets blocked, and that degrades every user of the feature.
- **A malformed address never reaches the network** — we do not use someone else's free RPC as a validator.

## Verification

Driven through `httpx.MockTransport` rather than patching internals, so real request building, status handling and JSON parsing all run. A test that patched `_read` would prove only that the wrapper returns what the mock said.

Controls that matter:

- a genuine `0x0` balance is `known` **and** `0`; an unreachable chain is `None` — the two cases that must never merge
- a `200` carrying an RPC `error` object is `unreachable`, not a balance
- a failure is **not** written into the success cache (or it would outlive the backoff and keep reporting failure after recovery)
- backoff is per-endpoint: a failing Ethereum RPC does not silence Solana
- `_result()` itself is asserted to drop any amount passed for a non-`known` state — without that, every honesty test above would be decorative
- the async tests were confirmed to genuinely execute by breaking one deliberately

4096 passed, `app/onchain.py` at 94%, total coverage 95.51%.

## Still open

`dv6` stays **open**. This is the reader. Reconciliation against recorded earnings — *"the service said it paid me, did it land?"* — and the UI come once tier 1 (#269/#270) has merged.

## Summary by CodeRabbit

- **New Features**
  - Added read-only balance lookups for Ethereum, Polygon, and Solana.
  - Supports checking individual addresses or multiple addresses concurrently.
  - Validates addresses before attempting lookups.
  - Reports clear results for valid balances, unavailable networks, unsupported chains, and invalid addresses.
  - Preserves exact decimal precision and distinguishes unavailable balances from zero.
  - Reuses successful results to improve response speed and reliability.

### DIFF
--- a/app/onchain.py
+++ b/app/onchain.py
@@ -91,8 +91,44 @@ _cache: dict[str, tuple[float, dict[str, Any]]] = {}
 _backoff_until: dict[str, float] = {}
 
 
+#: A ceiling so a long-lived process cannot grow the cache without bound even if
+#: every entry is somehow live. Far above any real fleet's address count.
+MAX_CACHE_ENTRIES = 512
+
+
 def _now() -> float:
     return time.monotonic()
+
+
+def _cache_get(key: str) -> dict[str, Any] | None:
+    """A COPY of the cached result, or None.
+
+    Returning the stored dict itself would let any caller that mutates what it
+    got corrupt every later read — and this module's entire contract is that
+    ``amount`` is None unless ``state`` is ``known``, so a mutated cache entry
+    is exactly the lie the module exists to prevent.
+    """
+    hit = _cache.get(key)
+    if hit and hit[0] > _now():
+        return dict(hit[1])
+    return None
+
+
+def _cache_put(key: str, value: dict[str, Any]) -> None:
+    now = _now()
+    # Expired entries are only ever noticed on read, so without this sweep they
+    # accumulate for the life of the process: one entry per address ever asked
+    # about, never released.
+    for k in [k for k, (expires, _) in _cache.items() if expires <= now]:
+        del _cache[k]
+    # Evict DOWN TO the ceiling, not by one. Dropping a single entry leaves an
+    # over-full cache over-full forever, which is how a "capped" cache quietly
+    # is not one.
+    if len(_cache) >= MAX_CACHE_ENTRIES:
+        by_expiry = sorted(_cache, key=lambda k: _cache[k][0])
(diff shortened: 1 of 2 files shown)


GeiserX:
Good catch, and right on all three counts. I verified each against the running code before changing anything:

| Input | Old behaviour |
|---|---|
| EVM result `10` (a JSON number) | `known`, **1.6E-17** — `int(str(10), 16)` is 16 |
| Solana `{"value": true}` | `known`, **1E-9 SOL** — `isinstance(True, int)` is `True` |
| Solana `{"value": -5000000000}` | `known`, **−5 SOL** |

Every one is a *confident wrong number reported as fact*, which is the single outcome this module exists to prevent — worse than admitting we could not read the chain.

Fixed as proposed: EVM must return a `0x`-prefixed hex **string**; Solana must return an object whose `value` is a non-negative `int`, checked with `type(...) is not int` since `bool` subclasses `int`. Everything else raises and surfaces as `unreachable`.

Six regression tests added, plus a control asserting the documented shapes still parse — otherwise a parser that rejected *everything* would satisfy them all.

PR #291, opened 2026-08-07
feat: a read-only API token, so a dashboard need not hold the keys

GeiserX:
## Found by actually integrating something

Wiring a Heimdall tile to a live install surfaced this. To display two numbers,
a user must hand over a credential that either controls their containers or
enrols workers:

| credential | what it grants |
|---|---|
| `CASHPILOT_ADMIN_API_KEY` | **owner** — deploy, stop, remove containers; read stored credentials |
| the fleet key | the **enrolment** credential — anything holding it can enrol a worker |

Neither is close to least privilege for showing a balance. And on the reference
install the admin key **is not set**, so in practice the documented programmatic
path was the fleet key.

## The refusal is the design, not the token

Adding a `reader` role is the easy half. **An allowlist fails silently, and it
fails in the direction of passing** — so the reader is *refused* by
`_require_auth_api`, the guard ~40 endpoints already share, and must be opted
into one endpoint at a time via `_require_reader`.

The consequence worth having: **an endpoint added next year, by someone who has
never read this file, denies the read-only token by default.** Forgetting to
think about it produces a 403, not a leak.

This mirrors an existing decision in the codebase — the fleet role deliberately
does not satisfy writer, for the same reason.

## What is opted in

Chosen from what the Home Assistant integration actually reads, not invented:

`/api/earnings/summary` · `/api/earnings/breakdown` · `/api/fleet/summary` ·
`/api/health/scores` · `/api/services/deployed`

Deliberately small: widening later is easy, narrowing breaks whoever wired a
dashboard against it. Note HA also performs control actions (`collect`,
`start`/`stop`/`restart`), so it still needs a stronger credential — this covers
its read path.

## Ambiguity resolves downward

The read-only key is checked **before** the admin key, on purpose. Setting both
to one value is a misconfiguration either way, but the resolutions are not
equally bad:

- resolving to **owner** silently hands container control to whatever was pasted
  into a dashboard widget;
- resolving to **reader** merely stops the admin path working — loud, harmless.

## Tests

The sweep walks **the app's own route table**, not a hand-kept list, so an
endpoint added later is covered the day it is added:

- **28** `/api/` GET endpoints outside the allowlist must all refuse the reader —
  including `/api/config`, `/api/credentials/health`, `/api/compose`;
- the **5** inside must all accept it (a refusal so broad it blocks the point is
  also a failure);
- a **negative control** asserts the sweep is not vacuous, so it cannot quietly
  degrade into proving nothing.

Both mutations behave:

| mutation | what fails |
|---|---|
| drop the reader refusal in `_require_auth_api` | the 28-endpoint sweep |
| check the token *after* the admin key | the ambiguity test |

Five existing test files stubbed `_require_auth_api` for these handlers and now
stub both guards. I found those by enumerating every file that stubs the guard
*and* calls one of the five handlers — my first pass fixed only the files a
truncated log had shown me, and missed two.

`ruff check` + `ruff format --check` clean · `node --check` · currency and fleet
staleness checks pass · **4385 passed, coverage 95.57%** (gate 90%).

Closes CashPilot-c99d.

## Summary by CodeRabbit

* **New Features**
  * Added a read-only API key for reporting dashboards and GET-only analytics.
  * Enabled read-only access to service, earnings, health, and fleet summary reports.
  * Preserved broader access for administrator and fleet credentials.

* **Security**
  * Read-only credentials are rejected from write, deployment, and owner-level operations.
  * Ambiguous credentials now resolve to the lower-privilege read-only role.

* **Documentation**
  * Documented read-only and administrator API key scopes and configuration options.

* **Tests**
  * Added coverage for credential recognition, access restrictions, and reporting endpoint permissions.

### DIFF
--- a/app/auth.py
+++ b/app/auth.py
@@ -159,6 +159,16 @@ def get_current_user(request: Request) -> dict[str, Any] | None:
     # Check Bearer token — admin key gets owner, fleet key gets fleet (limited)
     auth_header = request.headers.get("Authorization", "")
     if auth_header:
+        # Checked FIRST, before the admin key, deliberately. If an operator sets
+        # both variables to the same value that is a misconfiguration either way,
+        # and the two possible resolutions are not equally bad: resolving it to
+        # "owner" would silently hand full container control to whatever they
+        # pasted into a dashboard widget. Resolving it to "reader" merely stops
+        # the admin path working, which is loud and harmless. Fail toward less
+        # privilege.
+        readonly_key = os.getenv("CASHPILOT_READONLY_API_KEY", "")
+        if readonly_key and hmac.compare_digest(auth_header.encode(), f"Bearer {readonly_key}".encode()):
+            return {"uid": 0, "u": "readonly", "r": "reader"}
         admin_key = os.getenv("CASHPILOT_ADMIN_API_KEY", "")
         if admin_key and hmac.compare_digest(auth_header.encode(), f"Bearer {admin_key}".encode()):
             return {"uid": 0, "u": "api", "r": "owner"}
--- a/app/deps.py
+++ b/app/deps.py
@@ -24,6 +24,7 @@ __all__ = [
     "templates",
     "_login_redirect",
     "_require_auth_api",
+    "_require_reader",
     "_require_writer",
     "_require_owner",
     "_require_private_network",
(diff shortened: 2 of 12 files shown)


GeiserX:
@coderabbitai review

PR #293, opened 2026-08-07
docs(heimdall): the install recipe, from having actually installed it

GeiserX:
The submission notes were written before anyone had run this. Installing it on a live **Heimdall 2.8.1** corrected three claims and produced a recipe worth keeping.

## Three things this file got wrong

**The `appid` is not issued by Heimdall.** It is `sha1(lowercase app name)`:

```console
$ printf 'cashpilot' | sha1sum
5b7c085bf60518c8e7261473688a98200e60847b   # == app.json, exactly
```

Same relation holds for Grafana, Vaultwarden and Home Assistant. This matters because it means the package can be assembled and **tested before anything is sent upstream** — which is how the rest of this section came to exist.

**A hand-installed app persists.** Both paths Heimdall loads from are symlinks into the config bind mount, so it survives a container update — the opposite of what I had assumed.

**CashPilot now has a read-only token** (#291), so the note apologising for its absence is gone. That is the key this tile should use.

## The trap worth documenting

`ItemController::getStats` catches **every** exception and returns `{"status":"inactive","html":""}`. A broken app therefore looks merely *idle*, never broken. Verifying an install by reading the HTTP response alone reads a silent failure as a pass — the log has to be checked too. Mine was empty, which is the part that made the live result trustworthy.

## Smaller things that cost time

- The blades must sit **beside** the PHP class — the view namespace is registered to `app_path('SupportedApps')`. Enhanced apps ship blades, non-enhanced ones don't, so a file listing from the wrong app misleads.
- Ownership must be **numeric**: `chown abc:users` fails on an unRAID host because that user exists only inside the container.
- `sqlite3` is not in the container; use `php -r` with PDO.

## The rule, exercised against the installed class

| server says | Earnings | Running |
|---|---|---|
| `has_readings:true, total_adjusted:66.61, active_services:43` | `$66.61` | `43` |
| `has_readings:false, total_adjusted:0` | **—** | `43` |
| `active_services:null` | `$66.61` | **—** |
| `has_readings:true, total_adjusted:0, active_services:0` | `$0.00` | `0` |

The last row is the one that makes the table mean anything: it carries the **same `0`** as the second and renders **differently**, proving the dash follows `has_readings` rather than the number. Without it the table would also pass against a tile that simply dashed every zero.

Docs only — no code paths touched.

Closes CashPilot-76xy.

### DIFF
--- a/contrib/heimdall/README.md
+++ b/contrib/heimdall/README.md
@@ -50,8 +50,20 @@ are deleted. Steps 1–2 are a web form and cannot be scripted.
    | Icon | `docs/logo.svg` from this repository — the official mark, transparent, disc filling the frame edge to edge |
 
 2. Once the request exists, use the **Enhanced** download button on their site to
-   get the scaffold. It carries a generated `appid`, which is why the files here
-   do not include an `app.json` — that value must come from them.
+   get the scaffold.
+
+   An earlier version of this file said the scaffold's `appid` "must come from
+   them". **It does not.** The value is simply `sha1(lowercase app name)`:
+
+   ```console
+   $ printf 'cashpilot' | sha1sum
+   5b7c085bf60518c8e7261473688a98200e60847b
+   ```
+
+   That matches the `appid` in `app.json` exactly, and the same relation holds
+   for Grafana, Vaultwarden and Home Assistant. Knowing it means the whole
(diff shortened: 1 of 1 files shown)


GeiserX:
@coderabbitai review

PR #294, opened 2026-08-07
fix(catalog): say which ProxyBase holds the wallet

GeiserX:
The catalog ships **two separately deployable** ProxyBase services whose names differ by one word:

| slug | name | domain | wallet? |
|---|---|---|---|
| `proxybase` | ProxyBase | proxybase.org | no volumes at all |
| `proxybase-xyz` | ProxyBase Markets | proxybase.xyz | **generates one; the volume is the only copy of the key** |

They are genuinely different products — different domains, different container images (`proxybaseorg/peer-cli` vs `proxybasehq/proxybase-cli`), different payout models. Not a duplicate to delete.

But nothing at the point of *choosing between them* said which is which. Both descriptions led with what the service does; neither with the domain, and neither mentioned that one holds funds whose loss is permanent — a fact the catalog states plainly in its own `critical_volumes` text.

## This is not hypothetical

I conflated the two while reading the code and briefly concluded a live deployment was about to destroy a wallet. It was not: the running container is labelled `cashpilot.service=proxybase`, the entry with `volumes: []`, `MOUNTCOUNT=0`, and no such volume on the host.

**The label settles which catalog entry a container came from. The container name does not.**

If reading carefully lands there, someone scanning a catalog list can certainly get it wrong — and the cost of the wrong guess is irreversible.

## The test is general, not a special case

> Whenever one service's display name is a prefix of another's, and exactly one of the pair holds irreplaceable state, the one that does must say so.

A **negative control** asserts at least one such pair exists, so the invariant cannot quietly go vacuous if the catalog changes shape. Both assertions were mutation-checked and fail on different lines:

| mutation | fails at |
|---|---|
| revert the wallet mention | the "must mention it" assertion |
| make both descriptions identical | the "must not be interchangeable" assertion |

**Referral links untouched** — both keep their own `signup_url` and code.

4387 passed. Closes CashPilot-cp5g.

### DIFF
--- a/services/bandwidth/proxybase-xyz.yml
+++ b/services/bandwidth/proxybase-xyz.yml
@@ -10,7 +10,7 @@ description: >
   both direct (own bandwidth) and upstream (resell external proxies) modes with
   auto-reconnect. Buyers purchase proxy sessions by country, ISP, and bandwidth.
   Payments flow through the Micropayments Protocol (MPP) on Tempo Chain.
-short_description: SOCKS5 proxy marketplace for AI agents - USDC payout, CLI daemon
+short_description: proxybase.xyz - SOCKS5 marketplace for AI agents, USDC payout. Generates a wallet whose volume holds the only copy of the key
 
 referral:
   signup_url: "https://proxybase.xyz?referral=nXzS3c6iTO"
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

PR #295, opened 2026-08-07
feat: reconcile payouts against receipts, and make the wrong version unbuildable

GeiserX:
The tempting feature is to compare recorded earnings against the on-chain balance and flag a shortfall as *"this service is reporting money it is not sending"*.

**That inference does not hold.** A low or zero balance is equally consistent with the user having **withdrawn** the funds — which is the normal thing to do with money. Shipping it would produce confident accusations against services that paid correctly: the same class of error the whole wallet tier was built to avoid, one level up.

## So the module cannot see a balance

There is no parameter through which one could be passed, and a test asserts the signature stays that way:

```python
params = set(inspect.signature(pr.reconcile).parameters)
assert params == {"address", "observations", "receipts"}
```

That is a structural guarantee rather than a stylistic one. No caller can ask this module to compare against a balance, however much they might want to.

## Three states of `receipts`, not two

| value | meaning |
|---|---|
| `None` | we could not look — says **nothing** about the service |
| `[]` | we looked, and nothing has ever arrived |
| non-empty | money has arrived |

Collapsing the first into the second would turn *our own missing API key* into an accusation. It is the same absent-is-not-zero rule the codebase already runs on, applied to receipts instead of money.

The test for this is paired with its control: identical address, identical earnings history, **only** `receipts` changing from `None` to `[]` — and the two must produce different verdicts. Without the control, the first test would pass against a module that never accuses anyone at all.

## Exactly one of six verdicts accuses

`is_finding` is computed centrally, not by each caller, so **a verdict added later defaults to not being an accusation**. A test enumerates the verdicts from the module itself rather than a hand-kept list, so a new one is covered the day it is added.

It also refuses to speak too soon — fewer than 3 observations, a rise spanning under 7 days, flat earnings, a blip that reverted, or an unparseable date all return "not enough to say".

## Verification

Everything passed first time, so each guard was mutated:

| mutation | tests that fail |
|---|---|
| treat "not checked" as "nothing arrived" | both `AbsentIsNotEmpty` assertions |
| drop the minimum-span guard | the short-span and unparseable-date tests |
| let a second verdict accuse | both withdrawal tests + the enumeration test |

6 verdicts enumerated, so the enumeration test is not vacuous. **4401 passed**, ruff clean.

## Scope

Inert until the user supplies their own explorer key — keyless public JSON-RPC cannot return transfer history, and a shared key never ships. `NO_KEY` is an honest *"not checked"*, not a clean bill of health.

Part of CashPilot-l09j; the fetching half still needs a key.

### DIFF
--- a/app/payout_reconcile.py
+++ b/app/payout_reconcile.py
@@ -1,0 +1,186 @@
+"""Reconcile recorded earnings against payout RECEIPTS, never against a balance.
+
+CashPilot-l09j. This module exists mainly to make one specific wrong feature
+hard to build, so the reasoning is here rather than in a commit message.
+
+THE TEMPTING VERSION: compare recorded earnings against the on-chain balance
+and flag a shortfall as "this service is reporting money it is not sending".
+**That inference does not hold.** A low or zero balance is equally consistent
+with the user having withdrawn or moved the funds, which is the normal thing to
+do with money. Shipping it would produce confident accusations against services
+that paid correctly.
+
+So this module never looks at a balance at all. The only question it can answer
+is "did anything ever ARRIVE at this address, and when", and the only claim it
+will make is the one receipts actually support.
+
+The distinction the whole module turns on:
+
+    receipts is None  -> we could not look. Says NOTHING about the service.
+    receipts == []    -> we looked, and nothing has ever arrived.
+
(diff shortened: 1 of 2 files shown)


GeiserX:
@coderabbitai review

PR #297, opened 2026-08-07
fix: answer 401 before parsing an unauthenticated request body

GeiserX:
Guards were called *inside* handler bodies, so FastAPI validated the request body first and an unauthenticated caller got the schema back:

```
POST /api/config   malformed body, no auth -> 422
{"detail":[{"type":"missing","loc":["body","data"],"msg":"Field required","input":{}}]}

POST /api/config   valid body,     no auth -> 401
```

**Authentication held** — nothing read, nothing written. Confirmed on the shipped `1.22` image: a read-only token POSTing a distinctly-named key got 403 and that key was **absent** afterwards, while an admin-written key **was** present (so the read-back works and the absence means something).

What leaked was *structure*: the write surface could be enumerated and its schemas learned without credentials, and Pydantic parsed untrusted input before anyone was identified.

## The mechanism was verified, not assumed

Before touching anything, in an isolated app:

| guard style | malformed + no auth | valid + no auth |
|---|---|---|
| called in the body | **422** | 401 |
| `Depends()` | **401** | 401 |

## Scope: six endpoints

`deploy`, `compose`, `preferences`, `config`, and both password routes — every write endpoint that takes a Pydantic body **and** checks auth unconditionally at the top of the body.

**`/api/workers/{id}/command` is deliberately not converted.** It chooses its guard *from* the body — deploy requires owner, everything else writer — so the body must be parsed before the right guard is known. An earlier mechanical pass hoisted it anyway, deleted the conditional and left an empty `if`. That is why this is an explicit list rather than a pattern match, and the exception is asserted in the tests so it reads as a decision rather than an oversight.

I also over-counted at first: a loose regex matched `Request` as a Pydantic model and reported 16 candidates. The real number is 9, of which 6 are safely convertible.

## Two existing tests changed, both worth it

- The preference tests call the handler **directly**, bypassing FastAPI, so a `Depends` default never resolves. They now pass the user explicitly — honest, since the dependency is part of the signature.
- `test_the_server_really_does_require_owner` **asserted on source text**: it grepped `main.py` for `"_require_owner(request)"` near the decorator. It broke on a change that made the endpoint *strictly more secure* — the clearest possible demonstration that it was never testing behaviour. It now calls the endpoint.

## Tests

Every converted endpoint is checked both ways, and the second is the control: if an endpoint were simply broken or misrouted, the malformed case could return 401 while proving nothing about ordering. A **well-formed** unauthenticated request must be refused identically.

Mutating `/api/config` back to an in-body guard fails exactly two tests — and the control keeps passing, confirming it measures something different.

**4415 passed**, ruff clean. Closes CashPilot-nz4d.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -23,7 +23,7 @@ from typing import Any
 import httpx
 from apscheduler.events import EVENT_JOB_ERROR, EVENT_JOB_MISSED
 from apscheduler.schedulers.asyncio import AsyncIOScheduler
-from fastapi import FastAPI, HTTPException, Request
+from fastapi import Depends, FastAPI, HTTPException, Request
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse, PlainTextResponse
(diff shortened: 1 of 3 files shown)


GeiserX:
Correction: the commit and description reference `CashPilot-nz4d`. That ID does not exist — I wrote it without checking. The bead is **`CashPilot-gr02`**, now closed. Nothing about the change itself is affected; only the reference was wrong.

PR #304, opened 2026-08-07
feat(catalog): referral registry + self-audit — a bare link can no longer hide

GeiserX:
Closes the referral blind spot behind CashPilot-4jtc's story: ProxyBase's domain migration silently stripped attribution, and Bytebenefit sat wrongly dead with a bare URL for ~5 months of unattributed signups. Nothing but a human rereading YAML could catch either.

**The registry** — each `referral:` block may now carry:
- `code`: the referral code itself, proven **anchored** in `signup_url` (query value, bare query key like spide's `?f3bc51`, or path segment like `/ref/CODE` — a hostname coincidence such as `grass` in `app.grass.io` is not attribution).
- `program`: three-valued — `true` = provider verified to run a program, `false` = verified not to, absent = unknown. **Absent is not no**, and unknowns never fail: failing on them would teach people to write `program: false` without checking.

**The audit** (`scripts/referral_check.py`, wired into catalog-check):
- ERROR (always fails): code not anchored in its URL, empty code, code+`program: false`, non-boolean `program` (identity-checked — `1 == True` in Python would otherwise smuggle it through), non-mapping referral, code without signup_url. Checked for **every** status, so a dead service's code survives to any resurrection.
- ACTION (fails only `--strict`): active service, verified program, no code — currently exactly **repocket** (verified: $5 + 10% lifetime; needs the account holder to sign up).
- UNKNOWN (never fails): the 10-provider research backlog (bitping, flux, ionet, nosana, salad, deeper-network, helium, nodle, theta-edge, storj).

Registry contents: 26 codes verified anchored; golem/anyone-protocol/sentinel-dvpn `program: false` (no-account services per the repo's own docs); Bytebenefit's bonus terms fill the `service_detail.html` branch that read `referral.bonus` but had no data anywhere.

Tests: 20 in `tests/test_referral_check.py`, negative controls included (stripped code → red, hostname coincidence → red, garbage `program` values → red not silence, `--strict` pinned in both directions). Full suite 4454 passed. ruff + format clean.

## Summary by CodeRabbit

- **New Features**
  - Added referral codes, rewards, and program status information across numerous service listings.
  - Expanded referral documentation to clarify supported fields, program states, signup links, and bonuses.

- **Bug Fixes**
  - Added automated checks to identify invalid, incomplete, contradictory, or incorrectly linked referral information.

- **Tests**
  - Added comprehensive coverage for referral validation, malformed data, missing links, and strict checking behavior.

- **Chores**
  - Catalog validation now runs automatically when referral information changes.

### DIFF
--- a/.github/workflows/catalog-check.yml
+++ b/.github/workflows/catalog-check.yml
@@ -15,6 +15,7 @@ on:
       - "services/**"
       - "README.md"
       - "scripts/generate_readme_tables.py"
+      - "scripts/referral_check.py"
       - ".github/workflows/catalog-check.yml"
   push:
     branches: [main]
@@ -51,6 +52,16 @@ jobs:
 
       - name: README service tables match the catalog
         run: python scripts/generate_readme_tables.py --check
+
+      # Referral links are revenue. This fails on regressions (a recorded code
+      # no longer anchored in its signup_url, malformed registry facts);
+      # known-program-without-code gaps and unresearched providers are printed,
+      # not failed. A contributor who updates a migrated provider's signup_url
+      # CAN trip the code check -- the error text tells them the escape hatch:
+      # leave `code` as it was and flag it, since only the account holder can
+      # re-issue the link.
+      - name: Referral links still carry their codes
+        run: python scripts/referral_check.py
 
       - name: Every service has a guide
         run: |
--- a/scripts/referral_check.py
+++ b/scripts/referral_check.py
@@ -104,7 +104,13 @@ def audit(services: list[dict]) -> dict[str, list[str]]:
             continue
 
         if code is not None:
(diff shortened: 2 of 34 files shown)


GeiserX:
Addressed both actionable findings in the follow-up commit: `referral.code` must now BE a string (unquoted scalars are refused with the octal example in the message), and a missing `--services-dir` exits 2 instead of auditing zero files and reporting success. Both have tests.

On the third suggestion (put code-present + `program` absent services into the unknown bucket): declined deliberately — a recorded, anchored code is itself proof the provider runs a program, so there is nothing unknown left to research for that service. The linkage is documented in `services/_schema.yml` ("code presence means the provider has a program AND a code exists") and in the checker's docstring.

PR #315, opened 2026-08-08
fix(ci): the nightly live check never ran — a fabricated action pin killed it at setup

GeiserX:
You spotted [run 31242808822](https://github.com/GeiserX/CashPilot/actions/runs/31242808822). It is not provider drift — the job never started.

```
##[error]Unable to resolve action `actions/github-script@60a0d83039c74a4aa971cc2a0930ae7e2fe2c8bd`
```

That SHA does not exist. Beside the real one, the shape is unmistakable:

```
pinned:      60a0d83039c74a4a a971cc2a0930ae7e2fe2c8bd   # comment claimed v7.0.4
real v7.0.1: 60a0d83039c74a4a ee543508d2ffcb1c3799cdea
```

Same 16-char prefix, invented tail — and `v7.0.4` was never released (404). Written from memory rather than copied. It landed in #169, **the commit that created the workflow**, so this check has never run once; it has failed every night since 3 August.

**It was also silent.** The "open an issue" alarm was an `if: failure()` *step inside that job* — and a step cannot run when the job dies at setup. No `collector-drift` issue was ever filed. An alarm sharing a failure domain with the thing it watches is not an alarm.

### What changes
1. **The alarm becomes a separate job** (`needs: live`, `if: failure()`) so it fires however the watched job died, using the preinstalled `gh` CLI instead of an action — the reporter for "an action would not resolve" must not resolve actions. Least privilege too: the job handling real provider credentials drops to `contents: read`.
2. **`scripts/check_action_pins.py`** reads what a human cannot: fails on a ref that does not resolve, **and** on a `# vX.Y.Z` comment naming a missing tag or pointing at a different commit. Three-valued — a rate limit is `unknown` and never reddens the run, because an alarm that cries wolf is the failure mode this whole PR is about.
3. **`Action Pins` workflow** on PRs touching `.github/workflows/**`, plus weekly (a tag can move after merge). It depends on exactly one action, checkout.

### Verification
- Replayed the **original broken file** through the checker → exit 1, names the exact line.
- A real SHA with a lying comment (`v9.0.0` on the v7.1.0 commit) → exit 1.
- **46/46** references in the repo resolve today; the fabricated pin was the only bad one of 11 SHA pins.
- 18 offline tests for the checker; the pre-existing "something reacts to a failure" gate updated (it assumed the broken shape) and strengthened to forbid the reporter resolving non-checkout actions — **both mutations verified to fail it**.
- Full suite: 4522 passed.

Note: the live check will still report *"provider-drift detection is NOT active"* until a test carries `@pytest.mark.live` — that part is by design and now visible instead of hidden behind a job that never ran.

## Summary by CodeRabbit

* **New Features**
  * Added automated auditing to verify GitHub Actions are securely pinned and include matching version information.
  * Added scheduled and on-demand checks for workflow action pins.

* **Bug Fixes**
  * Improved live-check failure reporting, including failures during setup.
  * Restricted workflow permissions to the minimum required access.

* **Tests**
  * Added coverage for action-pin validation, workflow discovery, and failure-reporting behavior.

### DIFF
--- a/.github/workflows/action-pins.yml
+++ b/.github/workflows/action-pins.yml
@@ -1,0 +1,47 @@
+name: Action Pins
+
+# Every action this repo uses must exist, and its version comment must be true.
+#
+# `collector-live-check.yml` shipped a pin whose SHA shared sixteen characters
+# with a real release commit and then diverged into a commit that never
+# existed, beside a comment naming a tag that was never published. GitHub
+# resolves actions BEFORE the first step, so that job died in "Set up job"
+# every night for six nights without checking a single provider -- and its own
+# `if: failure()` alarm was a step inside that same job, so nothing was ever
+# reported. The pin was wrong in the commit that introduced the workflow, so
+# the check had never run once.
+#
+# A pinned SHA cannot be eyeballed. This is the thing that reads it.
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/**"
+  push:
+    branches: [main]
+    paths:
+      - ".github/workflows/**"
+  # A tag can be deleted or moved AFTER a pin is merged, so the audit is worth
+  # repeating on a schedule and not only at review time.
+  schedule:
+    - cron: "23 6 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  pins:
+    runs-on: ubuntu-latest
+    steps:
+      # The ONLY action this job uses. A guard against unresolvable actions
+      # that itself depended on several would be one bad pin away from the
+      # silence it exists to prevent; the checker is stdlib-only Python, and
+      # runners ship python3, so nothing else is needed.
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Every action reference resolves, and its version comment is true
+        env:
+          # Raises the API rate limit. Read-only; a fork PR's token works too.
+          GITHUB_TOKEN: ${{ github.token }}
+        run: python3 scripts/check_action_pins.py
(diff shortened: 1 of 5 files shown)


GeiserX:
Thanks — two of the four were right and are fixed in the follow-up commit; two don't hold, with evidence.

**Fixed — `check_action_pins.py` reporting a half-checked pin as OK.** Correct and the most important of the four: if the ref resolved but the version comment's tag lookup came back `UNKNOWN` (rate limit), the code fell through to `Finding(use, OK)`. That is this script telling exactly the kind of lie it exists to catch — "verified" for a claim nobody checked. Now `UNKNOWN`: still never a failure, but no longer counted as verified. Two tests added, including the control that a fully-resolved claim is still `OK` (otherwise the fix could pass by calling everything unknown).

**Fixed — forbid *all* actions in the reporter.** Right, and stronger than what I wrote. The reporter shells out to the preinstalled `gh` CLI and reads no repository file, so it needs no checkout at all; "only checkout allowed" left one more way for the alarm to die at setup. The gate now requires every reporter step to omit `uses`, and the job already had zero, so it passes today.

**Not applicable — issue-creation race.** Already prevented: the workflow carries `concurrency: {group: collector-live-check, cancel-in-progress: false}` (lines 21-23), so two runs cannot execute concurrently — the second queues. Both runs therefore cannot reach the "is there an open issue" lookup at the same time. Job-level concurrency would be redundant.

**Not applicable — `actions/checkout` checking out the PR merge commit.** True, and deliberate: this is a **public** repo on **GitHub-hosted, ephemeral** runners, where a `pull_request` token is read-only and there is no self-hosted fork-PR surface (the repo's CI policy mandates `ubuntu-latest` for public repos precisely for this). Running PR-authored code is inherent to any PR gate here — the existing `test` job already executes PR-authored pytest. The residual worry, "a PR could weaken the checker so a bad pin slips through," is true of every lint gate in every repo and is answered by reading the diff, not by pinning a ref. Pinning `github.event.pull_request.head.sha` would not change it either (it is *more* attacker-controlled, not less).

46/46 references still resolve; full suite 4524 passed.

PR #334, opened 2026-08-13
feat(health): the UI can now report that it cannot do its job

GeiserX:
## Why

The UI half of the healthcheck audit. Its container healthcheck was a **TCP connect** — completed by the kernel's listen backlog while uvicorn is wedged, the scheduler is dead or the database is unreadable — and `docs/guides/prometheus-metrics.md` claimed both images verify "the app inside is actually answering", which after #330 was only true of the worker. Separately, the bell's **"All collectors healthy" renders from a latch that can never go false** once set: after a scheduler death, every signal stays affirmatively green — green worker dots (a frozen status column), a healthy bell, week-old balances that look identical to fresh ones — indefinitely.

## What

- **`GET /api/health`** (unauthenticated — the Docker healthcheck has no credentials; registered in the anonymous-rejection test's `PUBLIC` list with its reason). Answers three concrete operational questions and 503s with fixed state names: `scheduler stopped`, `no collection completed in N minutes` (2.5× the interval — one slow run plus one missed slot of flap resistance), `database unreadable`. Exception text never reaches the unauthenticated body, and there is a test pinning that.
- **The staleness stamp measures completion, not success** — stamped in the collection `finally`. A wedged collection lock freezes it, because the skip path returns *before* the finally: the wedge that `"Collection already in progress, skipping"` used to disguise as routine is now exactly what this detects.
- **Dockerfile healthcheck** asks the endpoint via `http.client` with an explicit 200 check (`urlopen` honors `http_proxy` with no localhost exemption — the worker-side lesson from #330's review). Verified to parse intact in a real Docker build.
- **The bell stops lying**: when the server-computed `collection_stale` flag is set it renders "No collection in N minutes — status unknown" instead of affirming health from the permanent latch.

## Testing

- 9 new tests with negative controls: healthy + unauthenticated exact shape, each degradation cause, no-secrets-in-body, completed run advances the stamp vs **lock-skip does not**, fresh vs frozen staleness contract.
- Full suite **4673 passed**, coverage **96.56%**; ruff + format + `node --check` clean; the anonymous-rejection sweep (87 routes) passes with the one documented exemption.

## Summary by CodeRabbit

- **New Features**
  - Added an unauthenticated health endpoint reporting scheduler, collection, and database status.
  - Health checks now return clear degraded status details when services are unavailable.
  - Collector alerts now show collection age and identify stale results.

- **Bug Fixes**
  - The interface no longer reports all collectors as healthy when results are outdated.
  - Container health monitoring now verifies a successful HTTP health response.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,14 @@ VOLUME /data
 EXPOSE 8080
 
 ENTRYPOINT ["/entrypoint.sh"]
+# Ask the app whether it can still DO ITS JOB, not merely whether the port
+# answers: a TCP connect is completed by the kernel's listen backlog while
+# uvicorn is wedged, the scheduler is dead or the database is unreadable —
+# every internal failure rendered as "healthy". /api/health answers 503 for
+# those. http.client, NOT urllib.request: urlopen honors http_proxy (which
+# Docker injects fleet-wide when the host has a proxies config) with no
+# localhost exemption, so the "loopback" probe would be answered by the proxy.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=20s --retries=3 \
-    CMD ["python", "-c", "import socket; socket.create_connection(('127.0.0.1', 8080), timeout=3).close()"]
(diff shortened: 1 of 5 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #349, opened 2026-08-19
fix(alerts): a flapping phone still buzzed every cycle — cap pushes per worker

GeiserX:
The sustained-recovery damping merged this morning (#348) was defeated within hours, by the same phone, in the opposite direction: its normal duty cycle turned out to be **7–10 minutes awake, 2–3 minutes dozing**. Every wake window passed the ~4-minute sustained bar, re-armed the alert, and the next nap pushed again — eight Telegram messages in two hours with the damping active. A worker whose ordinary operation *is* rhythmic flapping can never be rate-limited by episode detection, whatever the threshold.

**The cap now sits on the notification itself.** After an offline push for a worker, further offline pushes for that worker wait out a 6-hour floor, independent of episodes. The first push for a genuinely dead server is still immediate. The dashboard and bell keep tracking every episode in real time — only the phone-buzz is capped (worst case 4/day/worker, down from ~100/day observed). Recovery deliberately never resets the floor, because resetting on recovery is exactly the re-arm loop this ends. A total delivery failure releases the floor, so the existing un-dedupe retry contract ("will retry on the next cycle") stays true.

Floor state is in-memory: a restart forgets it, which at worst costs one early push. `TestOfflinePushFloor` pins all four properties (suppressed-but-recorded inside the floor, expiry re-pushes, recovery never resets, failed delivery releases), mutation-verified — removing the gate turns the suite red.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -408,6 +408,12 @@ async def _push_alert(kind: str, subject: str, title: str, message: str, alert_i
         return
     metrics.record_notify_delivery(bool(delivered))
     if not delivered:
+        # The worker push floor is armed at spawn time, before the outcome is
+        # known — a TOTAL delivery failure must release it, or the un-dedupe
+        # retry below would be suppressed for the whole floor window and the
+        # "will retry on the next cycle" promise becomes false.
+        if kind == "worker":
+            _worker_last_offline_push.pop(subject, None)
         try:
             undeduped = await database.delete_alert(alert_id)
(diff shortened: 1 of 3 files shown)


GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

GeiserX:
@coderabbitai review

PR #360, opened 2026-09-19
fix(mysterium): let the catalog opt out of no-new-privileges

Mluo2011:
The Mysterium node brings up its wireguard interface by shelling out to `sudo ip address replace dev myst0 10.182.0.1/24`. `deploy_raw` sets `security_opt=["no-new-privileges:true"]` for every container, which permanently prevents `sudo` from elevating — so that call fails and **every session dies at setup**:

```
ERR Session failed, disconnecting error="cannot get provider config for session ...:
    could not start provider wg connection endpoint: could not configure device:
    failed to create TUN device: failed to assign IP address:
    \"ip address replace dev myst0 10.182.0.1/24\": exit status 1
    output: sudo: PERM_SUDOERS: setresuid(-1, 1, -1): Operation not permitted"
```

The node still registers, publishes proposals, appears in discovery and answers price queries — so it looks healthy from every angle while earning nothing. Same silent shape as the missing-TUN failure in CashPilot-6rv, but a different cause: `/dev/net/tun` was present and correct throughout.

## What this changes

Adds a `no_new_privileges` catalog key (default `true`), threaded through `DeploySpec` → `_validate_deploy_spec` → `deploy_raw`. Validated **per slug**, exactly like `network_mode: host`: a service can only disable the hardening for itself, and only by declaring it in its own YAML. Mysterium is the sole opt-out; everything else is unchanged.

## Docs correction

`docs/guides/mysterium.md` documented this exact symptom ("Running but not earning") but attributed it solely to a missing device, still passed `--security-opt no-new-privileges:true` in its recommended fix, and verified with:

```bash
docker exec cashpilot-mysterium sh -c 'ip tuntap add dev probe mode tun && echo TUN_OK ...'
```

That check **cannot detect this failure**. `docker exec` starts a fresh process which does not inherit the container's `no_new_privs` bit, so it prints `TUN_OK` on a container whose own node process cannot elevate. I confirmed that false positive on a broken container. The guide now checks `HostConfig.SecurityOpt` and documents both causes.

## Verification

Reproduced and fixed on a live residential node:

- **Before:** every inbound session failed at interface setup; zero sessions over several hours despite full-cone NAT, working UPnP, correct discovery listing and quality 2.4.
- **After:** a full session completed — `p2p-session-acknowledge`, six `p2p-payment-message` promises, clean `p2p-session-destroy`, and an `earnings_change` event.

Tests: `ruff check` and `ruff format --check` clean. Full suite 4593 passed vs 4585 on baseline (+8 new), with the same 203 pre-existing network-dependent failures in my environment.

`tests/test_catalog_no_new_privileges.py` covers the catalog declaration, per-slug scoping (another service cannot claim it), unknown slugs, and the hardened default. Verified non-vacuous: removing the catalog key fails 3 of the 8.

## Summary by CodeRabbit

- **New Features**
  - Containers now use the secure `no-new-privileges` setting by default.
  - Mysterium deployments can opt out to support WireGuard interface configuration.
  - Opt-outs are restricted to services explicitly approved for this setting; unauthorized requests are rejected.

- **Documentation**
  - Updated Mysterium troubleshooting guidance to cover privilege restrictions and TUN device configuration.
  - Added service configuration guidance for the new security setting.

- **Tests**
  - Added coverage for defaults, approved exceptions, and deployment validation.


Mluo2011:
Good catch — fixed in d1d5e55.

You are right that the slug authorises nothing on its own: it is caller-supplied, so `slug="mysterium"` plus any `image` plus `no_new_privileges=false` would have started an arbitrary image with `security_opt=[]`.

`_catalog_no_new_privileges_optout(slug)` now returns the catalog **image** the slug may drop the hardening for, and `_validate_deploy_spec` requires `spec.image` to match it:

```python
if not spec.no_new_privileges:
    allowed_image = _catalog_no_new_privileges_optout(slug)
    if not allowed_image:
        raise HTTPException(403, f"Disabling no-new-privileges is not allowed for {slug!r}")
    if _image_repository(spec.image) != allowed_image:
        raise HTTPException(403, ...)
```

Comparison is on the repository rather than the full reference — a tag or digest on the catalog image is still the same image, and a tag is not a security boundary — while `registry:5000/owner/img` is not mistaken for a tagged reference.

New tests cover the smuggled image, a lookalike repository (`evil/mysteriumnetwork/myst`), tag and digest forms of the real image, and the parser itself. Verified non-vacuous: reverting the image check fails exactly the two attack tests, and nothing else.

Suite is 4600 passed against 4585 on baseline (+15), `ruff check` and `ruff format --check` clean.

Worth noting the same slug-only pattern is used for `network_mode: host` a few lines below. The blast radius there is smaller — host networking on an arbitrary image is bad but not a privilege-escalation primitive — so I have left it alone rather than widen this PR. Happy to follow up separately if you want it tightened too.

GeiserX:
Thank you for this, @Mluo2011. The symptom report and the log line were exactly right, and they led to a fix that shipped in v1.36.6 (#361). You are credited in that commit and in the [v1.36.6 release notes](https://github.com/GeiserX/CashPilot/releases/tag/v1.36.6). I'm closing this PR because the cause turned out to be different, and I want to show you how I got there.

I ran the real image with the command as PID 1 on three hosts:

| Container settings | `NoNewPrivs` | `sudo ip address replace ...` |
|---|---|---|
| cap-drop ALL, NET_ADMIN, no-new-privileges (CashPilot before) | 1 | `setresuid(-1, 1, -1): Operation not permitted`, rc=1 |
| cap-drop ALL, NET_ADMIN (what this PR produces) | 0 | same error, rc=1 |
| cap-drop ALL, NET_ADMIN, SETUID, SETGID, no-new-privileges (v1.36.6) | 1 | rc=0 |
| default capabilities, no-new-privileges | 1 | rc=0 |

The node already runs as root, so `no_new_privs` takes nothing away from it. What sudo needs is `CAP_SETUID` and `CAP_SETGID`. CashPilot drops every capability and only added `NET_ADMIN` back. So the fix is two capabilities in the catalog entry, and `no-new-privileges` stays on for every container.

My guess is the container you tested your fix on was started without `--cap-drop ALL`, which makes it work for that reason. If you still have it, this would confirm it:

```bash
docker inspect <container> --format '{{.HostConfig.CapDrop}} {{.HostConfig.CapAdd}}'
```

One more thing I checked, because the guide change depended on it. `docker exec` does inherit both the capability set and the `no_new_privs` bit: `grep NoNewPrivs /proc/self/status` inside an exec shows 1. So `docker exec cashpilot-mysterium sudo -n true` is a valid one-line check for this failure, and the guide uses it now.

To pick up the fix, upgrade the UI and the worker to v1.36.7 and redeploy Mysterium from the dashboard. A worker on an older version refuses the new capabilities with a 403. v1.36.7 also makes a redeploy keep the data mount the running container has, so a node whose keystore sits in a bind mount keeps its identity. If sessions still fail for you after that, please open an issue with the log lines and I'll look at it.

PR #364, opened 2026-09-19
Clarify Docker platforms in traffmonetizer.md

Mluo2011:
Updated Docker configuration and documentation to clarify platform support and correct inaccurate information regarding the architecture of the Traffmonetizer Docker image.

Summary

Corrected inaccurate platform information in the CashPilot guide. The "arm64v8" Docker image/tag currently has amd64 architecture metadata on Docker Hub.

Type of Change

- [ ] New service (YAML + guide)
- [ ] New earnings collector
- [ ] Bug fix
- [ ] Feature
- [x] Documentation
- [ ] Other

Checklist

- [ ] "ruff check app/" passes
- [ ] "ruff format --check app/" passes
- [ ] Service YAML follows "services/_schema.yml"
- [x] Guide page updated in "docs/guides/"
- [ ] README service table updated (if new service)
- [x] No credentials or secrets committed

## Summary by CodeRabbit

* **Documentation**
  * Updated the Docker configuration guide to list only `linux/amd64` as the automatically supported platform.
  * Clarified Docker tag behavior: the default tag uses the AMD64 build, while ARM tags must be selected and run manually on ARM devices.


GeiserX:
Thanks, good catch. Docker Hub does label all three tags (`latest`, `arm64v8`, `arm32v7`) as `linux/amd64`, and the default tag is an amd64 build, so the guide was wrong to promise arm64.

One correction to the wording, pushed onto your branch: the `arm64v8` tag is not an amd64 package. I pulled the layers of each tag and read the ELF header of `usr/local/bin/cli`:

| tag | Hub label | binary |
|---|---|---|
| `latest` | linux/amd64 | x86_64 |
| `arm64v8` | linux/amd64 | aarch64 |
| `arm32v7` | linux/amd64 | arm32 |

So the ARM builds exist, they are just mislabelled, which means Docker will not select them for an ARM host and CashPilot (which deploys the default tag) gets the amd64 build. The guide now says that. The catalog entry still lists arm64 and its description mentions Raspberry Pi; that needs a fix on the deploy side and will be handled separately.

PR #366, opened 2026-09-20
fix(ui): pass the slug as a value, not an onclick argument list

### LINKED ISSUE #365: [Bug]: Logs / Restart / Stop buttons send "'slug', N" as the slug
### Description

Logs, Restart and Stop don't work on any service. Clicking Logs gives a modal
titled `Logs: 'bitping', 1` and "Error loading logs: Worker request failed".

```
GET /api/containers/'bitping',%201/logs?lines=200 "HTTP/1.1 404 Not Found"
```

It's asking for a container called `'bitping', 1`, quotes and all. The worker
itself is fine everything is all working.

Looks like it is leftover inline-onclick syntax in app.js. delegate.js reads each
`data-aN` as one plain string, but these six buttons still build it as
arguments (1195, 1198, 1201, 1270, 1273, 1276):

```js
const wParam = inst.worker_id != null ? `', ${inst.worker_id}` : `'`;
data-a1="'${escapeHtml(svc.slug)}${wParam}"
```

Every other `data-a1` in the file uses the plain form.

### Steps to Reproduce

1. Deploy any service
2. Click the Logs icon on its row
3. Modal title shows `Logs: 'bitping', 1`, body shows the error

### Expected Behavior

Logs open. Slug in `data-a1`, worker id in `data-a2`.

### Suggested fix

```js
const wParam = inst.worker_id != null ? ` data-a2="${inst.worker_id}"` : '';
data-a1="${escapeHtml(svc.slug)}"${wParam}
```

Keep `wParam` empty rather than `data-a2=""` when there's no worker — delegate.js
stops at the first undefined `data-aN`, so an empty string gets passed as a real
workerId. Patched my own 1.36.7 this way and logs, restart and stop all work.

### Deployment Method

Docker Compose

### CashPilot Version

1.36.7 (also in 1.36.5 and main)

Mluo2011:
Closes #365.

Logs, Restart and Stop are broken on every service. `data-a1` still carries inline-handler argument syntax from the delegate.js migration:

```js
const wParam = inst.worker_id != null ? `', ${inst.worker_id}` : `'`;
data-a1="'${escapeHtml(svc.slug)}${wParam}"
```

delegate.js reads each `data-aN` as one raw string, so the whole fragment arrives as `slug` and `workerId` is undefined:

```
GET /api/containers/'bitping',%201/logs?lines=200 "HTTP/1.1 404 Not Found"
worker proxy error (404): unstructured error body
```

Six call sites — logs/restart/stop on the main row and the per-node sub-rows. They were the only `data-a*` values with a leading quote.

`wParam` now emits the attribute instead, and stays empty when there is no worker rather than rendering `data-a2=""`: delegate.js stops collecting at the first undefined `data-aN`, and an empty string is not None to `workerId != null`.

Tests live in the class that already covers this failure mode (`data-a2="this"`). The whole-class regex matches `data-aN="'${` specifically — a bare `data-aN="'` is also legitimate string concatenation, as at app.js:1352 where the quote closes the JS string and the value follows.

Verified on a live 1.36.7 instance: modal title goes from `Logs: 'bitping', 1` to `Logs: bitping (worker #1)`, logs stream, restart and stop work.

Suite 4639 passed vs 4636 on baseline (+3 new), same 203 pre-existing failures in my environment. Reverting the fix fails 2 of the 3 new tests. `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed service action buttons so actions consistently receive the correct service and worker information.
  * Corrected actions for both standard service rows and multi-instance rows.
* **Tests**
  * Added coverage to verify action data is formatted correctly, including cases where no worker is assigned.


GeiserX:
Merged and shipped in v1.36.8, thank you. Verified on a live instance after the upgrade: the Logs button now sends the slug and worker id separately and the modal streams the container log. Note for anyone upgrading: app.js is served without a Cache-Control header, so a browser tab that loaded the dashboard before the upgrade can keep the old script for a while; a hard refresh picks up the fix.

